PR #1, opened 2018-11-21
CloudEvents OM/Extensions and JSON serializer baseline

clemensv:
Tracking PR for visibility - Don't merge just yet.

Signed-off-by: clemensv <clemensv@microsoft.com>


src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -0,0 +1,164 @@
...
+    public class CloudEvent
+    {
+        readonly IDictionary<string, object> attributes;
+
+        /// <summary>
+        /// Create a new CloudEvent instance.
+        /// </summary>
+        /// <param name="type">'type' of the CloudEvent</param>
+        /// <param name="source">'source' of the CloudEvent</param>
+        /// <param name="id">'id' of the CloudEvent</param>
```
matzew: that's the 0.2 naming of the attributes, right ?

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        /// <summary>
+        /// Create a new CloudEvent instance
+        /// </summary>
+        /// <param name="extensions">Extensions to be added to this CloudEvents</param>
+        internal CloudEvent(params ICloudEventExtension[] extensions)
+        {
+            attributes = new CloudEventAttributes(extensions);
+            SpecVersion = "0.1";
```
matzew: should this, with things like `id` (instead of `eventId`), be 0.2 ?

clemensv: Yes. I'll fix that :)

Issue #20, opened 2019-07-02
Add support/sample for ASP.NET Core


meteatamel:
As far as I can tell, the package does not support ASP.NET Core. Either add support for ASP.NET Core or add a sample showing how to parse for CloudEvents on the server side with ASP.NET Core

tanwarsatya:
It's easy to parse any string message to CloudEvent. Check the unit test project for other examples.

```
 var objjson = (JObject)JToken.Parse(message);

   // create cloud event object
   CloudEvent obj = new JsonEventFormatter().DecodeJObject(objjson);
```

meteatamel:
Thanks for pointing this out. I can parse CloudEvents now. I'll close the issue.

PR #21, opened 2019-07-04
Add aspnet core support

TobbenTM:
Hi!

I took the changes made by @friism, updated for 0.3 and added an input formatter for the MVC pipeline.

I believe this is a change that will make CloudEvents much easier to consume, instead of having to receive the wrong type only to deserialize yourself (as mentioned in #20).

I've outlined the new process in the README.md file.

Open to suggestions if you have any.

Hope this change is a welcome one, and can be merged. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,14 +224,14 @@ var cloudEvent = await HttpContext.Request.ReadCloudEventAsync();
 ### HTTP - ASP.NET Core MVC
 
 If you would like to deserialize CloudEvents in actions directly, you can register the
-`CloudEventInputFormatter` in the MVC options:
+`CloudEventJsonInputFormatter` in the MVC options:
 
 ```C#
 public void ConfigureServices(IServiceCollection services)
 {
     services.AddMvc(opts =>
     {
-        opts.InputFormatters.Insert(0, new CloudEventInputFormatter());
+        opts.InputFormatters.Insert(0, new CloudEventJsonInputFormatter());
     });
 }
 ```
--- a/README.md
+++ b/README.md
@@ -212,6 +212,42 @@ await context.Response.CopyFromAsync(cloudEvent,
 context.Response.StatusCode = (int)HttpStatusCode.OK; 
 ```
 
+### HTTP - Microsoft.AspNetCore.Http.HttpRequest
+
+On the server-side, you can extract a CloudEvent from the server-side `HttpRequest` 
+with the `ToCloudEvent()` extension.
+
+```C#
+var cloudEvent = HttpContext.Request.ToCloudEvent();
+``` 
+
+### HTTP - ASP.Net Core MVC
+
+If you would like to deserialize CloudEvents in actions directly, you can register the
+`CloudEventInputFormatter` in the MVC options:
+
+```C#
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddMvc(opts =>
+    {
+        opts.InputFormatters.Insert(0, new CloudEventInputFormatter());
+    });
+}
+```
+
+This formatter will only intercept parameters where CloudEvent is the expected type.
+
+You can then receive CloudEvent objects in controller actions:
+
+```C#
+[HttpPost("resource")]
+public IActionResult ReceiveCloudEvent([FromBody] CloudEvent cloudEvent)
+{
+    return Ok();
+}
+```
+
 ### AMQP
 
 The SDK provides extensions for the [AMQPNetLite](https://github.com/Azure/amqpnetlite) package. 
--- a/README.md
+++ b/README.md
@@ -215,13 +215,13 @@ context.Response.StatusCode = (int)HttpStatusCode.OK;
 ### HTTP - Microsoft.AspNetCore.Http.HttpRequest
 
 On the server-side, you can extract a CloudEvent from the server-side `HttpRequest` 
-with the `ToCloudEvent()` extension.
+with the `ReadCloudEventAsync()` extension.
 
 ```C#
-var cloudEvent = HttpContext.Request.ToCloudEvent();
+var cloudEvent = await HttpContext.Request.ReadCloudEventAsync();
 ``` 
 
-### HTTP - ASP.Net Core MVC
+### HTTP - ASP.NET Core MVC
 
 If you would like to deserialize CloudEvents in actions directly, you can register the
 `CloudEventInputFormatter` in the MVC options:
--- a/src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
+++ b/src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
@@ -35,9 +35,15 @@ namespace CloudNative.CloudEvents
 
             var request = context.HttpContext.Request;
 
-            var cloudEvent = request.ToCloudEvent(new JsonEventFormatter());
-
-            return InputFormatterResult.SuccessAsync(cloudEvent);
+            try
+            {
+                var cloudEvent = request.ToCloudEvent();
+                return InputFormatterResult.SuccessAsync(cloudEvent);
+            }
+            catch (Exception)
+            {
+                return InputFormatterResult.FailureAsync();
+            }
         }
 
         protected override bool CanReadType(Type type)
--- a/src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
+++ b/src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) Cloud Native Foundation. 
+// Licensed under the Apache 2.0 license.
+// See LICENSE file in the project root for full license information.
+
+namespace CloudNative.CloudEvents
+{
+    using System;
+    using System.Text;
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Mvc.Formatters;
+    using Microsoft.Net.Http.Headers;
+
+    public class CloudEventInputFormatter : TextInputFormatter
+    {
+        public CloudEventInputFormatter()
+        {
+            SupportedMediaTypes.Add(MediaTypeHeaderValue.Parse("application/json"));
+            SupportedMediaTypes.Add(MediaTypeHeaderValue.Parse("application/cloudevents"));
+
+            SupportedEncodings.Add(Encoding.UTF8);
+            SupportedEncodings.Add(Encoding.Unicode);
+        }
+
+        public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterContext context, Encoding encoding)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (encoding == null)
+            {
+                throw new ArgumentNullException(nameof(encoding));
+            }
+
+            var request = context.HttpContext.Request;
+
+            var cloudEvent = request.ToCloudEvent(new JsonEventFormatter());
+
+            return InputFormatterResult.SuccessAsync(cloudEvent);
+        }
+
+        protected override bool CanReadType(Type type)
+        {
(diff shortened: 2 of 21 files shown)


clemensv:
@tobbentm Please sign the PR and rebase. I will take the PR then

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
```diff
@@ -0,0 +1,129 @@
...
+                    {
+                        formatter = jsonFormatter;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("Unsupported CloudEvents encoding");
+                    }
+                }
+
+                return formatter.DecodeStructuredEvent(httpRequest.Body, extensions);
```
davidfowl: Why is this synchronous?

davidfowl:
cc @rynowak

TobbenTM:
> @TobbenTM Please sign the PR and rebase. I will take the PR then

@clemensv Sorry about the late reply here, been busy lately..

I'll take a look at this PR and update for 1.0, but @friism hasn't signed his commits, is that going to be a problem? Would he need to sign them?

friism:
@TobbenTM sorry about that, I updated the branch to have signed commits: https://github.com/friism/sdk-csharp/tree/add-aspnet-core-support

TobbenTM:
@friism Great! I'll try to rebase my changes onto your changes onto the new 1.0 changes! 😅

TobbenTM:
@clemensv PR rebased and all commits signed off! :) 

Let me know if any more adjustments are necessary.

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        /// <summary>
+        /// Converts this HTTP request into a CloudEvent object, with the given extensions,
+        /// overriding the formatter.
+        /// </summary>
+        /// <param name="httpRequest">HTTP request</param>
+        /// <param name="formatter"></param>
+        /// <param name="extensions">List of extension instances</param>
+        /// <returns>A CloudEvent instance or 'null' if the request message doesn't hold a CloudEvent</returns>
+        public static CloudEvent ToCloudEvent(this HttpRequest httpRequest,
```
davidfowl: This method shouldn't exist. We shouldn't have things that do IO but synchronous. Make it internal if you want to keep it in the formatter or remove it for now and expose the async version.

I'm also not a fan of the extension method in general but my care level is lower for that 😄. If we had to have something, I would mimic the form API:

```
public static ValueTask<CloudEvent> ReadCloudEventAsync(this HttpRequest httpRequest)
```

friism: @davidfowl roger, I can try and take a look. Re the extension method, are you unhappy with the method name or just the signature. I agree your name is probably better, but the library uses `ToCloudEvent()` for all the other extension method helpers that turns things into cloudevents, so I'm not sure it's worth breaking that convention

davidfowl: 1. Needs to be Async
2. It is worth breaking it because I'm assuming those are not doing IO (at least they shouldn't be)

clemensv:
@davidfowl are you guys done discussing? :)

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
TobbenTM: Re the second point; they're doing IO in the sense of reading from whatever stream they're given: https://github.com/cloudevents/sdk-csharp/blob/1b7f22f221ebc21c6030618ed17d1175b8c93ee8/src/CloudNative.CloudEvents/JsonEventFormatter.cs#L32-L33 
Whether that's a memory stream or something else is unknown tho.

TobbenTM:
@clemensv I agree the ASP.Net integration is something that might not fit perfectly with this library, if you'd rather have me publish my own Nuget package with the Input Formatter I can do that, and we could keep the HttpRequest extension here perhaps?

I've updated the PR with async decoding of the JObject, and bumped the sample project to 3.0, so everything works now. @davidfowl is aa78505 somewhere close to what you had in mind?

davidfowl:
> Generally, this addition is out on the far edge of what I would consider permissible for inclusion in this CNCF repo because it is a framework integration (AspNet) rather than a transport integration (System.Net). I would really prefer for that integration to happen on the AspNet side.

Meh, I don't agree. The GRPC ASP.NET Core integration is part of CNCF. Also ASP.NET Core ships in the box with .NET Core.

friism:
Random question: with the update to 3.0, should the library use `System.Text.Json` instead of NewtonSoft?

src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <PackageVersion>0.1</PackageVersion>
+    <Description>ASP.Net Core extensions for CloudNative.CloudEvents</Description>
+    <Copyright>Copyright Cloud Native Foundation</Copyright>
+    <RepositoryUrl>https://github.com/cloudevents/sdk-csharp</RepositoryUrl>
+    <PackageProjectUrl>https://cloudevents.io</PackageProjectUrl>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.2.2" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.2.5" />
```
rynowak: Not sure if you're just copying what other projects in this repo do - but 2.2 is not a great version to target.

2.1 is an LTS release and is the last release that support .NET Framework (supported for a while)
2.2 is end of life in 3 months
3.X was just released and 3.1 (in preview) will be supported for 3 years

So either this should target 2.1 and include all of the currently supported releases, or target 3.0 for the new hotness.

test/CloudNative.CloudEvents.IntegrationTests/AspNetCore/CloudEventControllerTests.cs
```diff
@@ -0,0 +1,41 @@
...
+                Id = Guid.NewGuid().ToString(),
+            };
+            var content = new CloudEventContent(cloudEvent, ContentMode.Structured, new JsonEventFormatter());
+
+            // Act
+            var result = await _factory.CreateClient().PostAsync("/api/events/receive", content);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, result.StatusCode);
+            Assert.Contains(cloudEvent.Id, await result.Content.ReadAsStringAsync());
```
rynowak: It would probably be valuable to verify some other aspect of the data you're sending across other than one of the simpler properties.

What could be happening in your case:
1. Your formatter doesn't run because the content type is `application/cloudevents+json` and your formatter is looking for `application/cloudevents`
2. The vanilla MVC JSON formatter runs and successfully deserializes 
3. You're verifying the `id` property that doesn't require any specialized logic in order to deserialize

It's possible based on my guess that you're formatter isn't running in this case but the test still passes.

README.md
```diff
@@ -212,6 +212,42 @@ await context.Response.CopyFromAsync(cloudEvent,
 context.Response.StatusCode = (int)HttpStatusCode.OK; 
 ```
 
+### HTTP - Microsoft.AspNetCore.Http.HttpRequest
+
+On the server-side, you can extract a CloudEvent from the server-side `HttpRequest` 
+with the `ToCloudEvent()` extension.
+
+```C#
+var cloudEvent = HttpContext.Request.ToCloudEvent();
```
rynowak: Looks like it's actually `ReadCloudEventAsync()`

README.md
```diff
@@ -212,6 +212,42 @@ await context.Response.CopyFromAsync(cloudEvent,
...
+``` 
+
+### HTTP - ASP.Net Core MVC
```
rynowak:
```suggestion
### HTTP - ASP.NET Core MVC
```

src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace CloudNative.CloudEvents
+{
+    using System;
+    using System.Text;
+    using System.Threading.Tasks;
+    using Microsoft.AspNetCore.Mvc.Formatters;
+    using Microsoft.Net.Http.Headers;
+
+    public class CloudEventInputFormatter : TextInputFormatter
```
rynowak: suggestion: should this be called `CloudEventJsonInputFormatter`?

src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
davidfowl: OR support both 2.1 and 3.0

rynowak:
> Random question: with the update to 3.0, should the library use System.Text.Json instead of NewtonSoft?

This is probably a broader question than just this PR, but I think it's a worthwhile discussion to have 👍 I see lots of code in this repo that's already using JSON.NET.

clemensv:
@davidfowl if it ships "in the box" with .NET, then it is indeed okay to include. What other projects do in tyeh CNCF context isn't necessarily relevant for CloudEvents since many projects have different principles and governance structures in CNCF.

clemensv:
@TobbenTM the appveyor file is here, so if you change the target, the work doesn't seem to be done until the build works. It is important that the core SDK stays on .NET Standard 2.0, though, because not everyone can jump to .NET Core 3.0. This needs to work for NETFX as well. 

I'm going to wait with merging this until I've got an ACK from all folks discussing here that this is good to go.

src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public CloudEventInputFormatter()
+        {
+            SupportedMediaTypes.Add(MediaTypeHeaderValue.Parse("application/json"));
+            SupportedMediaTypes.Add(MediaTypeHeaderValue.Parse("application/cloudevents"));
```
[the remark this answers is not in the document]
TobbenTM: Yeah, with the name change to CloudEventJsonInputFormatter, it's reasonable to only support `application/cloudevents+json`. I think the general idea is that more formats for cloudevents could be added (to the other request extensions at least), but since all of them only supports Json, we can too.

src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
TobbenTM: Changed target to latest 2.1 patch, didn't find an straight forward way to target multiple dependencies, you guys have any pointers for me? This works for 3.0 tho, so not sure if necessary.

src/CloudNative.CloudEvents.AspNetCore/CloudEventInputFormatter.cs
TobbenTM: Yes, renamed the formatter, and changed the supported media type to the more specific `application/cloudevents+json`.

TobbenTM:
@clemensv Yeah, the core SDK is staying at netstandard2.0, but in order to verify the 3.0 compatibility issues David brought up, the integration test and sample project are both targeting 3.0. The classlib is still targeting netstandard2.0. So the question is rather if we want to not test 3.0 compatibility, and downgrade the test and sample project to 2.2, or bump the appveyor image. I'm bumping the image for now, let me know if you have any thoughts.

Issue #15, opened 2019-06-21
Add link to the nuget package


rinormaloku:
Update the documentation to have a link to the nuget package.

jskeet:
There are now links to all the NuGet packages in the user guide: https://github.com/cloudevents/sdk-csharp/blob/master/docs/guide.md

(We may want to duplicate *some* of that content public on the main README, but I'd prefer it not to get too crowded.)

PR #45, opened 2020-02-26
Provide installation guidance

tomkerkhove:
Provide installation guidance and link to NuGet package.

Closes https://github.com/cloudevents/sdk-csharp/issues/15

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,14 @@
 The `CloudNative.CloudEvents` package provides utility methods and classes for creating, encoding,
 decoding, sending, and receiving [CNCF CloudEvents](https://github.com/cloudevents/spec).
 
+## Installation
+
+You can easily install the SDK via [NuGet](https://www.nuget.org/packages/CloudNative.CloudEvents/):
+
+```shell
+PM > Install-Package CloudNative.CloudEvents
+```
+
 ## A few gotchas highlighted for the impatient who don't usually read docs
 
 1. The [CloudEvent](src/CloudNative.CloudEvents/CloudEvent.cs) class is not meant to be used with 


jskeet:
@tomkerkhove: I don't know whether the DCO checker was added after you'd raised this PR, but could you amend your commit message to include a sign-off ([example](https://github.com/cloudevents/sdk-csharp/commit/0acb17529a419dcda5d72cfe5724b26423aa2e48))?

tomkerkhove:
Happy to change to dotnet approach if you prefer @jskeet?

It was not there yet I think but happy to change, but will wait for your feedback on the above

jskeet:
> Happy to change to dotnet approach if you prefer @jskeet?

I'd actually suggest not including any installation instructions - but possibly include the NuGet package name in the text so it's easy to copy/paste. So something like:

> The SDK is available via the [CloudNative.CloudEvents NuGet package](https://www.nuget.org/packages/CloudNative.CloudEvents/)

I'd hope that anyone wanting to use the SDK can either follow the instructions on the NuGet package page, or will already know how to install a NuGet package anyway.

(It's slightly annoying that it's natural to put the package name in the link text but that makes it slightly harder to copy/paste if you don't want to follow the link.)

I don't know whether the other packages (e.g. CloudNative.CloudEvents.AspNetCore) should also be listed - what do you think?

jskeet:
We've now included the package names in the [user guide](https://github.com/cloudevents/sdk-csharp/blob/master/docs/guide.md) which is linked from the top-level README. I'll close this PR as I don't think it's needed any more (and would certainly need rebasing) - but please don't let that deter you from further docs improvements :)

PR #60, opened 2020-06-25
More project file improvements

jskeet:
Most importantly, this introduces strong-naming, but it also makes the generated NuGet packages deterministic and SourceLink-aware.


jskeet:
(Approved offline by @clemensv.)

Issue #61, opened 2020-06-26
Http Header DateTime format should be in RFC 3339 format ot avoid errors in Knative Eventing 


BFrancisA:
CloudEvent header date-time format is not in RFC 3339 format.  There needs to be a T between the date and the time.

E.g.:  2020-06-17 12:52:43Z   Wrong!
         2020-06-17T12:52:43Z  Correct

Knative Eventing (broker filter)  rejects non RFC 3339 formatted  DateTime.

To fix:
In  sdk-csharp/src/CloudNative.CloudEvents/HttpClientExtension.cs
Function:  MapAttributesToListenerResponse()

Change line 471 from:
```
 httpListenerResponse.Headers.Add(HttpHeaderPrefix + attribute.Key,
                            ((DateTime)attribute.Value).ToString("u"));
```

To 
```
 httpListenerResponse.Headers.Add(HttpHeaderPrefix + attribute.Key,
                            XmlConvert.ToString((DateTime)attribute.Value, XmlDateTimeSerializationMode.Utc));
```

jskeet:
Related: #57

PR #63, opened 2020-07-14
Issue 61 changed date-time in CE header to be in RFC 3339 format.

BFrancisA:
Addresses issue #61.
CloudEvent header date-time format is not in RFC 3339 format. There needs to be a T between the date and the time.

E.g.: 2020-06-17 12:52:43Z Wrong!
2020-06-17T12:52:43Z Correct

Knative Eventing (broker filter) rejects non RFC 3339 formatted DateTime.


jskeet:
The change looks good to me, but could you add a test?

PR #68, opened 2020-08-03
Improvements to CloudEventAttributes consistency

jskeet:
- Allow every attribute *except* spec version to be cleared via the indexer (setting the value to null)
- Prevent null values being added with the Add method
- Avoid ever losing the spec version attribute

### DIFF
--- a/samples/Directory.Build.props
+++ b/samples/Directory.Build.props
@@ -1,8 +1,17 @@
 <Project>
   <PropertyGroup>
+    <!-- Make the repository root available for other properties -->
+    <RepoRoot>$([System.IO.Path]::GetDirectoryName($([MSBuild]::GetPathOfFileAbove('.gitignore', '$(MSBuildThisFileDirectory)'))))</RepoRoot>
+
     <!-- Disable deterministic source paths for sample projects -->
     <DeterministicSourcePaths>False</DeterministicSourcePaths>
 
+    <!-- Build properties -->
+    <AssemblyOriginatorKeyFile>$(RepoRoot)/CloudEventsSdk.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>True</SignAssembly>
+    <Deterministic>True</Deterministic>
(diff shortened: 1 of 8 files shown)


clemensv:
Protecting spec version is great.

"null" is equivalent to an optional attribute not being set. If an event arrives as such from across the wire and we validate it, it should not fail validation.

I am not principally opposed to never allowing null in the serialized CloudEvent, but we should raise this is in the group to get consensus on whether "null" is permissible to indicate absence of a value or whether the attribute key must be omitted outright. Since we have not made a hard rule about this, this might be a breaking change.

JSON.NET writes out null by default when the attribute is present, and I can see folks who build their own trivial implementations populate structured CloudEvents with something as simple as a string formatter where optional are filled out with null if that is what the parameter says. We would reject those events with this change and that makes interop more difficult.

https://www.newtonsoft.com/json/help/html/SerializationSettings.htm

jskeet:
It seems to me like if null is intended to be supported, there'd need to be an explicit representation of it in the spec in the type system. This would affect things like HTTP transport - how would a binary-mode HTTP request indicate a null value as differentiated from a string value of "null"?

Given that fully supporting null attribute values is a significant chunk of work, and there's no indication that they *are* supported, I think it's reasonable to assume they can be prohibited - but it would definitely be good to clarify this with the group and in the spec.

I've filed https://github.com/cloudevents/spec/issues/702 as a placeholder for discussion (or notes from online meetings).

jskeet:
Post-meeting comment:

I find the discussion in https://github.com/cloudevents/spec/issues/702 a bit confusing, but I do think it's reasonable (and in-keeping with my understanding of the discussion) to prohibit null values for attributes within an event

This PR needs conflict resolution (which I can sort out next week) but apart from that, are we now happy to merge?

JoshLove-msft (approved):

jskeet:
Rebased by removing the old first commit, which has been superceded. Will merge on green.

jskeet:
Turns out that first commit also gave InternalsVisibleTo access - adding that back now...

Issue #69, opened 2020-08-03
HTTP parsing does not URL-decode headers


jskeet:

jskeet:
Fixed in #70.

PR #70, opened 2020-08-03
Fix CloudEventContent's handling of headers

### LINKED ISSUE #67: CloudEventContent maps headers in structured mode
My understanding is that in structured mode, we shouldn't have ce-type (etc) HTTP headers, but I believe CloudEventContent still maps these.
Will write a test and then fix it, assuming I'm right.

jskeet:
- For structured mode, we shouldn't map attributes to HTTP headers
- For binary mode, we should default the media content type to application/json
- UrlEncode and UrlDecode headers in binary mode
- Adds more testing of headers

Fixes #67 and #69.

### DIFF
--- a/src/CloudNative.CloudEvents/CloudEventContent.cs
+++ b/src/CloudNative.CloudEvents/CloudEventContent.cs
@@ -9,8 +9,10 @@ namespace CloudNative.CloudEvents
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Net.Mime;
     using System.Text;
     using System.Threading.Tasks;
+    using System.Xml;
 
     /// <summary>
     /// This class is for use with `HttpClient` and constructs content and headers for
@@ -32,6 +34,8 @@ namespace CloudNative.CloudEvents
             if (contentMode == ContentMode.Structured)
             {
                 inner = new InnerByteArrayContent(formatter.EncodeStructuredEvent(cloudEvent, out var contentType));
+                // This is optional in the specification, but can be useful.
+                MapHeaders(cloudEvent, includeDataContentType: true);
                 Headers.ContentType = new MediaTypeHeaderValue(contentType.MediaType);
                 return;
             }
@@ -55,7 +59,7 @@ namespace CloudNative.CloudEvents
             }
 
             Headers.ContentType = new MediaTypeHeaderValue(cloudEvent.DataContentType?.MediaType ?? "application/json");
-            MapHeaders(cloudEvent);
+            MapHeaders(cloudEvent, includeDataContentType: false);
         }
 
         interface IInnerContent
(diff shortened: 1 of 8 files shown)


LXBdev:
According to the [spec](https://github.com/cloudevents/spec/blob/master/http-protocol-binding.md#323-metadata-headers), it is allowed to have the `ce-*`-Headers for structured content mode. However, the attributes are required in the content.

```
Implementations MAY include the same HTTP headers as defined for the binary mode.

All CloudEvents metadata attributes MUST be mapped into the payload, even if they are also mapped into HTTP headers.
```
The currently published SDK ignores the attributes if they are only mapped via the payload, even though they are valid according to the spec. It is also an impediment for me using the SDK in an API in a case where we don't get the `ce-*` headers.

@clemensv any ETA on when this PR will be merged?

clemensv:
Re: "For structured mode, we shouldn't map attributes to HTTP headers"

I implement the MAY rule and I believe having the metadata available for evaluation by proxies and HTTP service meshes is very useful without them having to consider the payload, especially with encodings they do not understand.

"For binary mode, we should default the media content type to application/json"

I believe we should just throw if we run into a situation where we don't have a content type set. Without further knowledge, that default has a good chance of being wrong, and the event will cause failures in subscribers as they pick the wrong decoder and that's a good source of blame games.

"UrlEncode and UrlDecode headers in binary mode"

Good fix.

"Adds more testing of headers"

Always good :)

jskeet:
> I believe we should just throw if we run into a situation where we don't have a content type set. 

Here's the bit of the spec I was working from:

> In some event formats the datacontenttype attribute MAY be omitted. For example, if a JSON format event has no datacontenttype attribute, then it is implied that the data is a JSON value conforming to the "application/json" media type. In other words: a JSON-format event with no datacontenttype is exactly equivalent to one with datacontenttype="application/json".
>
> When translating an event message with no datacontenttype attribute to a different format or protocol binding, the target datacontenttype SHOULD be set explicitly to the implied datacontenttype of the source.

Now it's not clear to me what is meant by "a JSON-format event". A `CloudEvent` object doesn't "have" a format. Should it be up to the formatter provided to the `CloudEventContent` constructor to determine "the format of the CloudEvent"? I know I'm sounding like a broken record, but I think we need to work on the spec being clearer on things like this.

jskeet:
@clemensv: I've added another commit to address the HTTP header mapping aspect. (I haven't changed the behavior for unknown datacontenttype yet.)

src/CloudNative.CloudEvents/CloudEventContent.cs
```diff
@@ -32,6 +34,8 @@ public CloudEventContent(CloudEvent cloudEvent, ContentMode contentMode, ICloudE
             if (contentMode == ContentMode.Structured)
             {
                 inner = new InnerByteArrayContent(formatter.EncodeStructuredEvent(cloudEvent, out var contentType));
+                // This is optional in the specification, but can be useful.
+                MapHeaders(cloudEvent, includeDataContentType: true);
                 Headers.ContentType = new MediaTypeHeaderValue(contentType.MediaType);
```
JoshLove-msft, line 39: Not related to these changes, but rather than checking
```
if (cloudEvent.Data is byte[])
```
should we check for IEnumerable<byte>? Also what about ReadOnlyMemory<byte>?

src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <PackageVersion>1.0</PackageVersion>
     <Description>CNCF CloudEvents SDK</Description>
+    <LangVersion>8.0</LangVersion>
```
JoshLove-msft, line 7: Hmm.. why was this needed?

src/CloudNative.CloudEvents/CloudEventContent.cs
```diff
@@ -75,32 +78,40 @@ protected override bool TryComputeLength(out long length)
...
+                    string headerValue = attribute.Value switch
                     {
-                        Headers.Add("ce-" + attribute.Key,
-                            WebUtility.UrlEncode(
-                                Encoding.UTF8.GetString(jsonFormatter.EncodeAttribute(cloudEvent.SpecVersion, attribute.Key, attribute.Value,
-                                    cloudEvent.Extensions.Values))));
-                    }
+                        string text => WebUtility.UrlEncode(text),
+                        ContentType contentType => contentType.ToString(),
+                        DateTime dt => XmlConvert.ToString(dt, XmlDateTimeSerializationMode.Utc),
```
JoshLove-msft, line 108: Out of curiosity, why do we use XmlConvert here?

JoshLove-msft (approved):

JoshLove-msft (commented):
Actually, I guess this is still pending the datacontenttype changes.

src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
jskeet: Because I'm using a switch expression. It doesn't change execution-time requirements at all.

src/CloudNative.CloudEvents/CloudEventContent.cs
jskeet: Because it performs exactly the formatting we need it to, and that's the simplest way of doing so. (It's been a while since I looked at this, but I don't think there's a very simple date/time format that does what we need.)

src/CloudNative.CloudEvents/CloudEventContent.cs
jskeet: Basically that all comes back to the question of how we handle data. I'd like to get this (and some other minor fixes) in before we tackle that, if possible.

jskeet:
> Actually, I guess this is still pending the datacontenttype changes.

Possibly, although I think I'd argue that that's also wrapped up in making the role of a CloudEvent formatter clearer, and data handling in general.

JoshLove-msft:
> > Actually, I guess this is still pending the datacontenttype changes.
> 
> Possibly, although I think I'd argue that that's also wrapped up in making the role of a CloudEvent formatter clearer, and data handling in general.

I will defer to @clemensv on this one.

jskeet:
Okay, post-meeting comment:

- I'm happy to remove the "defaulting to application/json" for the moment, but will *at least* add a test to show what happens if we don't specify it at the moment, and *could* throw an ArgumentException instead (please let me know which you'd prefer)
- We'll need more refactoring in the long term, but we have #71 to cover that

Beyond that, are we happy to merge?

JoshLove-msft:
> Okay, post-meeting comment:
> 
> * I'm happy to remove the "defaulting to application/json" for the moment, but will _at least_ add a test to show what happens if we don't specify it at the moment, and _could_ throw an ArgumentException instead (please let me know which you'd prefer)
> * We'll need more refactoring in the long term, but we have #71 to cover that
> 
> Beyond that, are we happy to merge?

Sounds good to me!

JoshLove-msft (approved):

jskeet:
I've added a new test class for CloudEventContent - far more tests would be useful, but it's a start. We now validate that (for binary mode) there *is* a data content type attribute, and provide a useful error message otherwise.

jskeet:
Hmm... now a test is failing. I'm unclear how it worked before. Will look into it.

jskeet:
> Hmm... now a test is failing. I'm unclear how it worked before. Will look into it.

It's fine - this was a test that I changed, which had been relying on the defaulting to application/json. Fixed.

PR #87, opened 2021-01-28
Make the spec version of an event immutable

### LINKED ISSUE #65: CloudEventAttributes.specVersion is only used in the constructor
Given how often the spec version of the attributes is used, it makes sense to keep it as a separate field (as it is now), but it must be kept up-to-date.

### LINKED ISSUE #66: Consider making the spec version immutable for any given event
I suspect it's quite rare to want to change a CloudEvent from one spec version to another - and being able to mutate an existing object in that way feels particularly odd, and leads to code complexity.

I propose that we prevent the CloudEvent spec version attribute from being modified (or cleared) from CloudEventAttributes, and instead provide a method that returns a *new* CloudEvent (or CloudEventAttributes) with equivalent attributes, but in a different version.

jskeet:
This commit includes two WithSpecVersion methods, one in CloudEvent
and the other in CloudEventAttributes, which convert from one
version to another. These are currently internal, but we can expose
them later if we wish.

Fixes #65
Fixed #66

Creating this commit has raised more issues to discuss:

- CloudEventSpecVersion.Default is a public enum value; changing that
  later would be a breaking change in difficult-to-document ways.
- In general, CloudEventSpecVersion feels like it deserves to be a
  class with well-known specific instances, rather than an enum. That
  would make various things much simpler.
- Given that attributes other than data have a limited set of types,
  I suspect it's worth having a CloudEventAttribute type encapsulating
  that.

### DIFF
--- a/src/CloudNative.CloudEvents/CloudEvent.cs
+++ b/src/CloudNative.CloudEvents/CloudEvent.cs
@@ -71,8 +71,13 @@ namespace CloudNative.CloudEvents
         /// <param name="specVersion">CloudEvents specification version</param>
         /// <param name="extensions">Extensions to be added to this CloudEvents</param>
         public CloudEvent(CloudEventsSpecVersion specVersion, IEnumerable<ICloudEventExtension> extensions)
+            : this(new CloudEventAttributes(specVersion, extensions), extensions)
         {
-            attributes = new CloudEventAttributes(specVersion, extensions);
+        }
+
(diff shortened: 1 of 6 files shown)


JoshLove-msft (approved):

jskeet:
Thanks for the quick review - I'll try to do some prototyping around the bullet points in the comments, for us to discuss tomorrow.

Issue #59, opened 2020-06-25
CloudEvent should enforce lower-case attribute names


jskeet:
The spec states:

> CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be descriptive and terse and SHOULD NOT exceed 20 characters in length.

Unfortunately, this can easily be violated:

```csharp
var evt = new CloudEvent(CloudEventsSpecVersion.V1_0, "my-type", new Uri("//my-source"), "my-id", DateTime.UtcNow);
evt.GetAttributes()["Subject"] = "my-subject";

var formatter = new JsonEventFormatter();
var bytes = formatter.EncodeStructuredEvent(evt, out _);
var text = Encoding.UTF8.GetString(bytes);
// Includes "Subject" instead of "subject"
Console.WriteLine(text);
```

I ran into this because the `HttpRequest.ReadCloudEventAsync` extension method in the ASP.NET Core package preserves header casing - so a header of "Ce-Type" ends up with a value of "Type" in the JSON.

I'm ambivalent as to whether we lower-case in the attributes, or reject an attempt to set an invalid attribute name, but we should definitely do one of the two.

jskeet:
Note: spec version 0.1 allowed non-lower-case-only names, but it's been "lower case only" since 0.2.

So this is related to #56. If we can at least drop 0.1, we can go case-sensitive all the way.

jskeet:
Fixes as of #92.

PR #88, opened 2021-01-29
Validate attribute names

JoshLove-msft:
Resolves https://github.com/cloudevents/sdk-csharp/issues/59

### DIFF
--- a/src/CloudNative.CloudEvents/CloudEventAttributes.cs
+++ b/src/CloudNative.CloudEvents/CloudEventAttributes.cs
@@ -7,6 +7,8 @@ namespace CloudNative.CloudEvents
     using System;
     using System.Collections;
     using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
     using System.Net.Mime;
 
     /// <summary>
@@ -95,9 +97,9 @@ namespace CloudNative.CloudEvents
                 if (!dict.TryGetValue(key, out var result))
                 {
                     return null;
-                }             
+                }
                 return result;
-            } 
+            }
             set
             {
                 // Allow the "setting" of the spec version so long as it doesn't actually modify anything.
@@ -135,7 +137,7 @@ namespace CloudNative.CloudEvents
 
         public static string DataSchemaAttributeName(CloudEventsSpecVersion version = CloudEventsSpecVersion.Default)
         {
-            return version == CloudEventsSpecVersion.V0_1 ? "schemaUrl" : 
+            return version == CloudEventsSpecVersion.V0_1 ? "schemaUrl" :
                    (version == CloudEventsSpecVersion.V0_2 || version == CloudEventsSpecVersion.V0_3 ? "schemaurl" : "dataschema");
         }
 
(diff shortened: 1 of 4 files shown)


src/CloudNative.CloudEvents/CloudEventAttributes.cs
```diff
@@ -369,5 +372,14 @@ private bool ValidateAndNormalize(string key, ref object value)
 
             return false;
         }
+
+        private void ValidateKey(string key)
+        {
+            if (key.All(c => char.IsDigit(c) || char.IsLower(c)))
```
jskeet, line 378: I believe this would allow non-ASCII characters. I think it's simplest to use `(c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')`

jskeet (commented):
Are we happy not lower-casing anything in the various event receivers? Normally HTTP headers are case-insensitive, so I'd expect CE-SUBJECT to still be valid for example, and end up as the "subject" context attribute.

The CloudEvent HTTP binding specification says "Note: per the HTTP specification, header names are case-insensitive" but it doesn't actually call out the lower-casing, unfortunately :(

JoshLove-msft:
> Are we happy not lower-casing anything in the various event receivers? Normally HTTP headers are case-insensitive, so I'd expect CE-SUBJECT to still be valid for example, and end up as the "subject" context attribute.

Well "Subject" would still be valid because it will hit the condition checking for the SubjectAttribute which ignores case. When would CE be prepended? Is that in binary mode? I wonder if we should special case CE for the known attributes.

jskeet:
> Well "Subject" would still be valid because it will hit the condition checking for the SubjectAttribute which ignores case. When would CE be prepended? Is that in binary mode? I wonder if we should special case CE for the known attributes.

I was assuming that we would make all of these case-sensitive. The spec is clear that attribute names are case-sensitive and need to be in lower-case - it's HTTP that's case-insensitive, so it's the HTTP transformation code that should be handling that, in my view... and I believe it should lower-case all header names beginning with "ce-". (We have far, far too much code that goes through all known attributes, handling them separately. I really want to get away from that, and treat everything the same.)

JoshLove-msft:
> > Well "Subject" would still be valid because it will hit the condition checking for the SubjectAttribute which ignores case. When would CE be prepended? Is that in binary mode? I wonder if we should special case CE for the known attributes.
> 
> I was assuming that we would make all of these case-sensitive. The spec is clear that attribute names are case-sensitive and need to be in lower-case - it's HTTP that's case-insensitive, so it's the HTTP transformation code that should be handling that, in my view... and I believe it should lower-case all header names beginning with "ce-". (We have far, far too much code that goes through all known attributes, handling them separately. I really want to get away from that, and treat everything the same.)

So just to make sure I understand, we want to validate that the CloudEventAttributes contain only lower case attribute names, even if they are the official CloudEvent attribute names, like subject. If they are not lower case, we should throw. On the receiving side, when dealing with HTTP extensions, we want to lowercase headers that start with CE-.

jskeet:
> So just to make sure I understand, we want to validate that the CloudEventAttributes contain only lower case attribute names, even if they are the official CloudEvent attribute names, like subject. If they are not lower case, we should throw. On the receiving side, when dealing with HTTP extensions, we want to lowercase headers that start with CE-.

Exactly. It's worth noting that this will mean we're "breaking" the spec for V0.1 and V0.2 (as I believe it was 0.3 that became case-sensitive) so we should remove all notion of support for those pretty soon :)

We'll need to check the various other transports as well in terms of their case handling.

I'm fine if you want to split this across multiple commits, btw - my *guess* is that the current tests would all pass anyway because we probably don't even try using Subject instead of subject.

JoshLove-msft:
Closing this out per discussion.

Issue #42, opened 2020-02-11
Serialize json using System.Text.Json


brechtvhb:
Are there any plants to use system.text.json as serializer instead of newtonsoft?

If you have an application that uses system.text.json it hurts to add a dependency to newtonsoft.json when using cloud events.

tomkerkhove:
Heh, I would have same scenario after this would be implemented as we're using newtonsoft.

Does it still hurt so much to add another dependency today? 🤷‍♂

brechtvhb:
As System.Text.Json is faster than newtonsoft & built into .net 3.0 it might make sense.
Hey lets all use EDI flatfiles because why would we start using json.

tomkerkhove:
I'm fine with both packages frankly but majority still uses newtonsoft given everything is depending on it, for now.

RascoApps:
it's now the time to change it and support both?

jskeet:
Given that users of either one probably don't want the dependency on the other, it might make sense to split them out into separate packages: CloudNative.CloudEvents.NewtonsoftJson and CloudNative.CloudEvents.SystemTextJson.

RascoApps:
Agreed.

Send from iPhone
________________________________

rynowak:
> Given that users of either one probably don't want the dependency on the other, it might make sense to split them out into separate package

+1 - this was our calculus for not using the CloudEvents package in https://github.com/dapr/dotnet-sdk/ - if we have an option to use this package without JSON.NET as a dependency, we definitely would.

jskeet:
@rynowak: Glad to hear that. I've just merged the change to split them out, but there's still a lot of work to do before 2.0. Once we're at the stage where we've at least got stable betas (and docs!), it would be great to hear more feedback about whether we're going in the right direction in general.

PR #92, opened 2021-02-05
Overhaul of codebase with new approach to attributes

jskeet:
This touches almost file in the repository, because it's hard not to
with such a fundamental change.

There are *lots* of TODOs now, all of which we should chase down
before release. In most cases this isn't a new problem - it's just
one that wasn't noted before,

The two main changes are:

- Spec versions, attributes, and attributes types are now first-class
  types. Rather than being able to set attributes to any value you
  want, values have to be of the right type and have a valid value
  (e.g. for datacontenttype). This centralizes value conversion
  for the most part (although transport/format-specific conversions
  may still be required) and makes everything more predictable.
- Extensions are now *not* first-class types; instead, the extension
  attributes can be registered with the CloudEvent in order to perform
  appropriate validation etc, and a pattern around extensions
  providing extension methods on CloudEvent allows them to be used
  easily.

This work has revealed a lot of smaller problems, such as the HTTP
transports propagating a content of "null" (rather than no content)
when no data was set in the CloudEvent. More testing is certainly
required.

It's worth noting that this is a really big breaking change. The 2.0
release was already going to be breaking, but this is much larger
than the previous changes. I assert that the benefits outweight the
costs though, and even though the package already has many users,
we hope there will be many times more users in the future.

Proposed process for this PR:

- This initial commit does not include nearly as many unit tests for
  CloudEvent, CloudEventAttribute or CloudEventAttributeType as I'd
  like. Those will come in a follow-up commit (which should be
  reviewed before merging) but don't affect the review of the overall
  ideas.
- If the general thrust of this PR is acceptable, then I propose we
  merge it, only making changes to the most egregious problems, and
  leaving further clean-up work (of which there's likely to be a lot)
  to later PRs. It's easier to create a clean-up PR in Visual Studio
  than to suggest fixes via GitHub.

Style changes:

- Use explicit access modifiers
- Move using directives before the namespace declaration
- Unit tests are generally more separated out to correspond to
  production classes, and are in separate directories/namespaces
  by project instead of all at the top level.

Currently StyleCop is not installed, but we should probably either
do that or use regular Roslyn analyzers in the long term - along
with turning on XML documentation generation.


src/CloudNative.CloudEvents/CloudEventAttribute.cs
```diff
@@ -0,0 +1,133 @@
...
+        public bool IsRequired { get; }
+
+        /// <summary>
+        /// Indicates whether this attribute is an extension attribute.
+        /// </summary>
+        public bool IsExtension { get; }
+
+        private Action<object> validator;
+
+        // TODO: Have a "mode" of Required/Optional/Extension?
```
JoshLove-msft, line 44: Wouldn't an extension always be optional though?

src/CloudNative.CloudEvents/CloudEventAttribute.cs
```diff
@@ -0,0 +1,133 @@
...
+        private CloudEventAttribute(string name, CloudEventAttributeType type, bool required, bool extension, Action<object> validator) =>
+            (Name, Type, IsRequired, IsExtension, this.validator) = (ValidateName(name), type, required, extension, validator);
+
+        internal static CloudEventAttribute CreateRequired(string name, CloudEventAttributeType type, Action<object> validator) =>
+            new CloudEventAttribute(name, type, required: true, extension: false, validator: validator);
+
+        internal static CloudEventAttribute CreateOptional(string name, CloudEventAttributeType type, Action<object> validator) =>
+            new CloudEventAttribute(name, type, required: false, extension: false, validator: validator);
+
+        public static CloudEventAttribute CreateExtension(string name, CloudEventAttributeType type) =>
```
JoshLove-msft, line 55: nit: ref docs

src/CloudNative.CloudEvents/CloudEventAttribute.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <param name="type"></param>
+        /// <param name="validator">Validator to use when parsing or formatting values.
+        /// This is only ever called with a non-null value which can be cast to the attribute type's corresponding
+        /// CLR type. If the validator throws any exception, it is wrapped in an ArgumentException containing the
+        /// attribute details.</param>
+        /// <returns></returns>
+        public static CloudEventAttribute CreateExtension(string name, CloudEventAttributeType type, Action<object> validator) =>
+            new CloudEventAttribute(name, type, required: false, extension: true, validator: validator);
+
+        public override string ToString() => Name;
```
JoshLove-msft, line 71: Should we add docs?

src/CloudNative.CloudEvents/Http/CloudEventHttpContent.cs
```diff
@@ -39,6 +36,8 @@ public CloudEventContent(CloudEvent cloudEvent, ContentMode contentMode, ICloudE
                 return;
             }
 
+            // TODO: Shouldn't we use the formatter in all cases? If I have a JSON formatter and
+            // If we specify that the the data is a byte array, I'd expect to end up with a base64-encoded representation...
```
JoshLove-msft, line 40: I agree with this line of thinking. Seems better to avoid doing magic under the hood.

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
```diff
@@ -45,90 +22,54 @@ public static class HttpRequestExtension
...
-                    }
-                    else
-                    {
-                        throw new InvalidOperationException("Unsupported CloudEvents encoding");
-                    }
-                }
-
-                return await formatter.DecodeStructuredEventAsync(httpRequest.Body, extensions);
+                // TODO: Handle formatter being null
+                return await formatter.DecodeStructuredEventAsync(httpRequest.Body, extensionAttributes).ConfigureAwait(false);
```
JoshLove-msft, line 31: Do we not want to support jsonFormatter as a default anymore?

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
```diff
@@ -1,15 +1,146 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Mime;
+using static CloudNative.CloudEvents.CloudEventAttribute;
+
 namespace CloudNative.CloudEvents
 {
-    public enum CloudEventsSpecVersion
+    // TODO: Rename to CloudEventSpecVersion? Using the plural form feels a little odd.
```
JoshLove-msft, line 13: Agreed

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
```diff
@@ -45,90 +22,54 @@ public static class HttpRequestExtension
...
+                    version = CloudEventsSpecVersion.FromVersionId(versionId);
                 }
 
-                if (httpRequest.Headers[SpecVersionHttpHeader2] != StringValues.Empty)
+                var cloudEvent = new CloudEvent(version, extensionAttributes);
+                foreach (var header in headers)
                 {
-                    switch (httpRequest.Headers[SpecVersionHttpHeader2])
+                    string attributeName = HttpUtilities.GetAttributeNameFromHeaderName(header.Key);
+                    if (attributeName is null || attributeName == CloudEventsSpecVersion.SpecVersionAttribute.Name)
```
JoshLove-msft, line 47: So if ce-specversion isn't present, but specversion is we would just ignore it?

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
```diff
@@ -1,15 +1,146 @@
...
-        Default = V1_0
+        /// <summary>
+        /// The attribute used to indicate the version of the CloudEvents specification being used.
+        /// </summary>
+        public static CloudEventAttribute SpecVersionAttribute { get; } = CreateRequired("specversion", CloudEventAttributeType.String, NonEmptyString);
+
+        // Populated by the constructor.
+        private static readonly List<CloudEventsSpecVersion> allVersions = new List<CloudEventsSpecVersion>();
+
+        public static CloudEventsSpecVersion Default => V1_0;
```
JoshLove-msft, line 29: Would Default only update for new major spec versions (along with a new major SDK version)?

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
```diff
@@ -1,15 +1,146 @@
...
+
+        public static CloudEventsSpecVersion V1_0 { get; } = new CloudEventsSpecVersion(
+            "1.0",
+            CreateRequired("id", CloudEventAttributeType.String, NonEmptyString),
```
JoshLove-msft, line 33: nit: maybe calculate the validator based on the attribute type within `CreateRequired`.

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
```diff
@@ -1,15 +1,146 @@
...
+            CreateRequired("type", CloudEventAttributeType.String, NonEmptyString),
+            CreateOptional("datacontenttype", CloudEventAttributeType.String, Rfc2046String), // TODO: Do we want some way of adding validation that this is per RFC 2046?
+            CreateOptional("dataschema", CloudEventAttributeType.Uri, NonEmptyUri),
+            CreateOptional("subject", CloudEventAttributeType.String, NonEmptyString),
+            CreateOptional("time", CloudEventAttributeType.Timestamp, null));
+
+        /// <summary>
+        /// The ID of the spec version, in its canonical serialized form.
+        /// </summary>
+        public string VersionId { get; }
```
JoshLove-msft, line 44: Would it be worth giving an example in the docs?

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
```diff
@@ -1,90 +1,96 @@
...
                 {
-                    var buffer = new MemoryStream();
-                    ((Stream)cloudEvent.Data).CopyTo(buffer);
-                    this.BodySection = new Data { Binary = buffer.ToArray() };
-                }
+                    Uri uri => uri.ToString(),
+                    // AMQPNetLite doesn't support DateTimeOffset values, so convert to UTC.
+                    // That means we can't roundtrip events with non-UTC timestamps, but that's not awful.
+                    DateTimeOffset dto => dto.UtcDateTime,
+                    _ => pair.Value
```
JoshLove-msft, line 63: It looks like previously the default case would result in inserting a new `Map` as the value in the outer `Map` with all of the non-uri/DTO values. Is this no longer equivalent?

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
```diff
@@ -1,90 +1,96 @@
...
-                }
+                case null:
+                    return null;
+                case byte[] bytes:
+                    return new Data { Binary = bytes };
+                case MemoryStream memoryStream:
+                    // Note: this will return the whole stream, regardless of position...
+                    return new Data { Binary = memoryStream.ToArray() };
+                case Stream stream:
+                    var buffer = new MemoryStream();
```
JoshLove-msft, line 87: Should the MemoryStream be disposed?

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
```diff
@@ -1,90 +1,96 @@
...
-                        {
-                            dict.Add(descriptor.Name, descriptor.GetValue(attribute.Value));
-                        }
-
-                        this.ApplicationProperties.Map.Add("cloudEvents:" + attribute.Key, dict);
-                    }
-                }
+                case null:
+                    return null;
+                case byte[] bytes:
```
JoshLove-msft, line 81: Do we need handling for `ReadOnlyMemory<byte>`?

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
```diff
@@ -45,90 +22,54 @@ public static class HttpRequestExtension
         /// <param name="extensions">List of extension instances</param>
```
JoshLove-msft, line 22: nit: docs for formatter param

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
```diff
@@ -45,90 +22,54 @@ public static class HttpRequestExtension
...
-                var attributes = cloudEvent.GetAttributes();
-                foreach (var httpRequestHeader in httpRequest.Headers.Keys)
+                cloudEvent.DataContentType = httpRequest.ContentType;
+                if (httpRequest.Body is Stream body)
                 {
-                    if (httpRequestHeader.Equals(SpecVersionHttpHeader1,
-                            StringComparison.InvariantCultureIgnoreCase) ||
-                        httpRequestHeader.Equals(SpecVersionHttpHeader2, StringComparison.InvariantCultureIgnoreCase))
+                    // TODO: This is a bit ugly.
+                    var memoryStream = new MemoryStream();
```
JoshLove-msft, line 60: Need to dispose?

src/CloudNative.CloudEvents/ICloudEventFormatter.cs
```diff
@@ -21,21 +20,26 @@ public interface ICloudEventFormatter
...
+        // TODO: Remove either this one or the stream one? It seems unnecessary to have both.
+
         /// <summary>
         /// Decode a structured event from a byte array
         /// </summary>
         /// <param name="data"></param>
         /// <param name="extensions"></param>
         /// <returns></returns>
-        CloudEvent DecodeStructuredEvent(byte[] data, IEnumerable<ICloudEventExtension> extensions);
+        CloudEvent DecodeStructuredEvent(byte[] data, IEnumerable<CloudEventAttribute> extensionAttributes);
```
JoshLove-msft, line 41: Should these use ROM<byte> instead of byte array?

src/CloudNative.CloudEvents/Http/HttpListenerExtensions.cs
```diff
@@ -0,0 +1,162 @@
...
+
+                // TODO: Check that this doesn't come through as a header already
+                cloudEvent.DataContentType = httpListenerRequest.ContentType;
+
+                // TODO: This is a bit ugly.
+                var memoryStream = new MemoryStream();
+                httpListenerRequest.InputStream.CopyTo(memoryStream);
+                if (memoryStream.Length != 0)
+                {
+                    cloudEvent.Data = formatter.DecodeData(memoryStream.ToArray(), cloudEvent.DataContentType);
```
JoshLove-msft, line 138: If DecodeData is made to take ROM we can avoid the ToArray and instead use something like:
```cs
memoryStream.GetBuffer().AsMemory(0, (int)memoryStream.Position)
```

src/CloudNative.CloudEvents/Http/HttpUtilities.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CloudNative.CloudEvents.Http
+{
+    /// <summary>
+    /// Common functionality used by all Http code.
+    /// TODO: Can we make this internal again? It's useful in the ASP.NET Core package, but I don't really like having it public...
```
JoshLove-msft, line 16: Agree with questioning why this is public - especially the Encode/DecodeHeader methods which don't have anything to do with CloudEvents. I can see the usefulness of GetAttributeNameFromHeaderName though.

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -1,123 +1,202 @@
...
-        /// format and encoding might differ from that of the chosen event format.
+        /// Sets or fetches the value associated with the given attribute name.
+        /// Setting a value of null removes the value from the event, if it exists.
+        /// If the attribute is not known in this event, fetching the value always returns null, and
+        /// setting the value add a new extension attribute with the given name, and a type of string.
+        /// (The value for an unknown attribute must be a string or null.)
         /// </summary>
-        /// <see cref="https://github.com/cloudevents/spec/blob/master/spec.md#contenttype"/>
-        public ContentType DataContentType
+        public object this[string attributeName]
```
JoshLove-msft, line 145: It seems like this may not be very discoverable to users. Have you considered a dictionary property or even having CloudEvent implement IDictionary?

src/CloudNative.CloudEvents/Extensions/Sampling.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the Apache 2.0 license.
+// See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace CloudNative.CloudEvents.Extensions
+{
+    public static class Sampling
```
JoshLove-msft, line 11: nice - I like the approach of using extension methods for these.

src/CloudNative.CloudEvents/Extensions/Sampling.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public static CloudEventAttribute SampledRateAttribute { get; } =
```
JoshLove-msft, line 13: Does this need to be public?

src/CloudNative.CloudEvents/JsonEventFormatter.cs
```diff
@@ -1,123 +1,117 @@
...
+
+                // TODO: This currently performs more conversions than it really should, in the cause of simplicity.
+                // We basically need a matrix of "attribute type vs token type" but that's rather complicated.
+
+                string attributeValue = value.Type switch
+                {
+                    JTokenType.String => (string)value,
+                    JTokenType.Date => CloudEventAttributeType.Timestamp.Format((DateTimeOffset)value),
+                    JTokenType.Uri => CloudEventAttributeType.UriReference.Format((Uri)value),
+                    JTokenType.Null => null, // TODO: Check we want to do this. It's a bit weird.
```
JoshLove-msft, line 109: Null values are not allowed per the spec, but do we want to be more permissive when decoding?

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -1,123 +1,202 @@
...
+                // TODO: Decide on spec version behavior
+
+                // TODO: Are we happy to add the extension in even if the value is null?
+                if (knownAttribute is null)
+                {
+                    if (value is object && !(value is string))
+                    {
+                        throw new ArgumentException($"Cannot assign value of type {value.GetType()} to unknown attribute '{attributeName}'");
+                    }
+                    knownAttribute = CloudEventAttribute.CreateExtension(attributeName, CloudEventAttributeType.String);
```
JoshLove-msft, line 163: Why is this always a string? If the user hasn't registered the attribute, we just assume string for simplicity and safety?

jskeet:
Added tests in a new commit - although the test failure in CI is alarming. Will look at that. Next I'll respond to all the comments (which will probably involve another commit).

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
jskeet: Hmm... I think you're right, but I suspect that that was *either* unintentional, *or* it was a hangover from when CloudEvent attributes could contain maps in JSON format. I don't think we want it, either way. (CloudEvent attributes are very much single-valued now.)

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
jskeet: That would require a dependency that I'd rather not take if we can avoid it. But we should definitely think about that when we revisit the question of how we want to handle data in general. (A lot of the data-handling code should be centralized I suspect, in the same way that the attribute-handling code is centralized. There's no need to have multiple copies of "copy a stream into a byte array" code for example.) Definitely raise `ReadOnlyMemory<byte>` again later though :)

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
jskeet: No need - it wouldn't actually do anything.

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
jskeet: Done as you raised it - but in general I haven't fixed up all the docs. That will be a future pass - possibly the next PR, possibly a little later.

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
jskeet: Nope - I want to move the JSON event formatter out to a separate package, so we can have one based on System.Text.Json and one based on Json.NET. See https://github.com/cloudevents/sdk-csharp/issues/42

(In general I'd like to try to limit the JSON-centricity of the codebase. The spec is already a bit overly-JSON-centric for my liking, but we can try to reduce the blast radius of that.)

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
jskeet: Yes, absolutely - we should ignore all headers that don't start with "ce-". (This isn't a change - although what *has* changed is that we don't use "ce-cloudeventsversion" which was a 0.1/0.2 header.)

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtension.cs
jskeet: Nope. (I'll leave any similar comments later. If you'd prefer that we *do* dispose, we can discuss that in a meeting. Once it's all centralized it would be a trivial change.)

src/CloudNative.CloudEvents/CloudEventAttribute.cs
jskeet: Yes - it would just be potentially simpler to have a single "mode" instead of two separate (but not entirely orthogonal) boolean values.

I'm definitely happy to wait and see whether it turns out it would actually be useful.

src/CloudNative.CloudEvents/CloudEventAttribute.cs
jskeet: Eventually, yes :)

I tend to find that it's worth leaving "relatively trivial" docs until after we're happy with the design of the code itself... whereas docs that describe non-obvious behavior, e.g. in the CloudEvent indexer, are useful in terms of the mental process. I'm not very consistent though!

Fortunately when we turn on "generate XML docs" we'll get a huge load of warnings for undocumented bits. I definitely, definitely want everything properly documented before we release 2.0. (That means conceptual docs as well as API reference docs, too - even if it's just a directory on GitHub with Markdown files.)

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
jskeet: Cool. I'll do that that in a follow-up PR, in a single commit, rather than doing it now - otherwise it will disrupt our review discussion. Have made a note for myself.

(Annoyingly, I can see how the plural form is more formally correct here. I hate it when we have to decide between "formally correct" or "feels better"...)

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
jskeet: Probably? :)

It's really hard to know how all the versioning stuff will pan out when we basically don't have any data points. (I don't think we can infer much from what happened pre-1.0, as obviously more attention will be paid to compatibility post 1.0.)

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
jskeet: No, the type validates the content separately. This is additional semantic validation. So empty strings are generally fine as context attributes, but they're *not* fine for the id attribute.

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
jskeet: Done.

src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <Description>CNCF CloudEvents SDK</Description>
     <LangVersion>8.0</LangVersion>
+    <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
```
jskeet: Removed this now - it was added while I was experimenting with StyleCop, but we can work out how to enforce styles later.

src/CloudNative.CloudEvents/Extensions/Sampling.cs
jskeet: I think it's potentially useful for consumers - it would feel odd exposing AllAttributes but not the individual ones.

src/CloudNative.CloudEvents/Extensions/Sampling.cs
jskeet: Thanks. I think we'll want to document this pattern carefully once we're really happy with it. There may well be some subtleties to think about in order to make it consumer-friendly when handling a CloudEvent created without the extensions known beforehand.

src/CloudNative.CloudEvents/Http/CloudEventHttpContent.cs
jskeet: After my frustrations with Uri over the last day or two, I'm even more convinced of that :)

I suspect we want to abstract out two concepts:

- "I know how to convert data to/from bytes, based on a certain set of types"
- "I know how to encode/decode structured events"

It gets a bit tricky in the middle, where the structured event encoding may or may not need a data converter.

src/CloudNative.CloudEvents/Http/HttpListenerExtensions.cs
jskeet: Interesting. We should definitely discuss the pros and cons of taking a dependency on System.Memory. I know there are some dependencies like this that cause problems (see https://blog.marcgravell.com/2020/01/net-core-net-5-exodus-of-net-framework.html) but maybe System.Memory itself is okay.

src/CloudNative.CloudEvents/Http/HttpUtilities.cs
jskeet: It wasn't public until the ASP.NET Core package came along.

There's a lot of repetition in terms of how conversion to/from events is done, and I'm wondering whether there's a way of abstracting it out so that all our HTTP conversions (and possibly even MQTT, AMQP and Kafka) can just provide a generic adapter. I don't want to experiment with that until the core is more solid though :)

The header encoding is *a bit* CloudEvent specific, in that we need to follow the exact steps specified in the CloudEvents spec, which may not be the same set of requirements as other things. You'd expect it to be simple and handled by a single Framework method call, but no...

src/CloudNative.CloudEvents/ICloudEventFormatter.cs
jskeet: (Ack, need to have this discussion.)

jskeet (commented):
Thanks for the review, Josh. I've added another commit with a couple of changes - I think everything else is either okay, or needs further discussion in person.

jskeet:
(Belatedly marked as ready for review.)

src/CloudNative.CloudEvents/CloudEventAttribute.cs
JoshLove-msft: Yeah I don't have a strong preference here. Maybe we could make it more clear that required and optional are both for reserved attributes. RequiredReserved, OptionalReserved, Extension?

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
JoshLove-msft: but doesn't calling CreateRequired already specify that the attribute can't be empty?

src/CloudNative.CloudEvents.Amqp/AmqpCloudEventMessage.cs
JoshLove-msft: fair enough

src/CloudNative.CloudEvents/Http/HttpListenerExtensions.cs
JoshLove-msft: Sure we can discuss it - I think it is pretty widely used in newer APIs.

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -131,6 +161,7 @@ public CloudEvent(CloudEventsSpecVersion specVersion, IEnumerable<CloudEventAttr
                     return;
                 }
                 // TODO: We could convert the attribute value here instead? Or is that a bit too much "magic"?
+                // TODO: Should we validate the value against the
```
JoshLove-msft, line 164: You trailed off there 😄

JoshLove-msft (approved):

src/CloudNative.CloudEvents/CloudEventsSpecVersion.cs
jskeet: Oh, I see what you mean. I don't think there's anything technically stopping a required attribute from being empty - it's just that all the current required attributes also have a non-empty requirement :)

Given how few callers there will be to CreateRequired (it's internal) I think I'd prefer to be explicit here.

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -1,123 +1,202 @@
...
+            // unhelpful...
+            get => attributeValues.GetValueOrDefault(Preconditions.CheckNotNull(attribute, nameof(attribute)).Name);
+            set
             {
-                foreach (var extension in extensions)
+                Preconditions.CheckNotNull(attribute, nameof(attribute));
+                string name = attribute.Name;
+                var knownAttribute = GetAttribute(name); 
+
+                // TODO: Are we happy to add the extension in even if the value is null?
```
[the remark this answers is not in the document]
jskeet: Sorry, this wasn't shown when I was going through the review before.
Yes, that's the case - and I've left it in. I've still left the TODO for the moment so we can check it again once we've had the new approach in place for a little while... I think it'll become clearer whether our decisions make sense when I've written some tutorial-style documentation.

src/CloudNative.CloudEvents/CloudEvent.cs
jskeet: Whoops, yes. I think that was because I stopped writing the comment and implemented validation instead :) Deleted in a new commit.

src/CloudNative.CloudEvents/CloudEvent.cs
jskeet: (Again, I missed this before - apologies. I really wish GitHub had a simple "go to next comment" option that would include diffs that it wasn't showing by default.)

Implementing IDictionary could be tricky - it's not obvious what the Count would be, for example. We've also got the issue that we've got two indexers - one for string and one for CloudEventAttribute, and we probably don't want to implement different generic versions of IDictionary.

I'm happy to consider this again later - I'll add a TODO at the top of the file.

src/CloudNative.CloudEvents/CloudEvent.cs
jskeet: Yes. Every CloudEvent attribute type has a canonical string representation, so this feels like a reasonable balance between safety and convenience. (An alternative would be to reject it entirely - which would be a reasonable choice, forcing callers to create the extension attribute themselves and use the other indexer if they wanted to.) Worth revisiting after writing conceptual docs.

src/CloudNative.CloudEvents/CloudEventAttribute.cs
jskeet: Hmm... maybe. Let's work out naming if and when we decide that the enum is worth doing :)

jskeet (commented):
Thanks Josh - will merge on green, then get started on documentation, which may well flush out some more interesting issues.

Issue #55, opened 2020-06-16
Clarify CloudEvent formatter responsibilities


jskeet:
I'm writing a doc to help myself think about this more clearly, but it feels to me like there's a lack of clarity around when data should be converted in what exact way.
We may want to change the actual code to make this clearer, or it may just be a matter of fixing docs. We'll see.

One key result I'd like to see: code that says (for example) "Populate HttpRequestMessage from CloudEvent" then "Convert HttpRequestMessage to CloudEvent" should be able to roundtrip in both directions, without gaining extra levels of encoding on each cycle, which I suspect it would do now without care.

jskeet:
My current thinking is along the lines of:

- Event formats are *only* relevant for structured content modes. They shouldn't be required when dealing with binary content modes.
- For binary content modes, it may be reasonable to say that we *require* the `Data` property to be one of a few specific types, e.g. `ReadOnlyMemory<byte>` or `byte[]`, and when decoding we always end up with a `ReadOnlyMemory<byte>` as the `Data` in the resulting property
- For structured content mode, the event formatter is responsible for defining what combination of CLR types for `Data` and `datacontenttype` attributes are valid when encoding, along with working out what data type to use when decoding
- We should strongly discourage any event formatters from creating a `CloudEvent` with a `Data` property value of type `Stream`... it's really fragile in terms of ownership.
- As discussed on a previous call, let's make `CloudEventFormatter` an abstract class instead of an interface.

@JoshLove-msft: If this sounds right to you, I might start work on this next week.

jskeet:
Additionally, we *may* want to remove the `ContentMode` enum - it's not clear that it's particularly useful.

jskeet:
Note: it might be useful to get event formatters involved after all, for consistency when decoding.

JoshLove-msft:
> My current thinking is along the lines of:
> 
> * Event formats are _only_ relevant for structured content modes. They shouldn't be required when dealing with binary content modes.
> * For binary content modes, it may be reasonable to say that we _require_ the `Data` property to be one of a few specific types, e.g. `ReadOnlyMemory<byte>` or `byte[]`, and when decoding we always end up with a `ReadOnlyMemory<byte>` as the `Data` in the resulting property
> * For structured content mode, the event formatter is responsible for defining what combination of CLR types for `Data` and `datacontenttype` attributes are valid when encoding, along with working out what data type to use when decoding
[...]

PR #97, opened 2021-02-17
Revisiting event formatter responsibilities

jskeet:
Towards #55.

It's worth reviewing this one commit at a time - I'm afraid the last one is unavoidably large :(
Once this is merged, I want to add a *lot* more tests for the JsonEventFormatter, to validate everything it now documents. We could potentially do that in this PR, but I suspect it'll simplify reviewing if we focus on what the responsibilities should be for now, and then focus on the implementation and tests in another PR. (I've tried to make the implementation follow the docs, of course, but I definitely need tests...)

### DIFF
--- a/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
@@ -14,74 +14,130 @@ using System.Threading.Tasks;
 namespace CloudNative.CloudEvents.NewtonsoftJson
 {
     /// <summary>
-    /// Formatter that implements the JSON Event Format
+    /// Formatter that implements the JSON Event Format.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// When encoding CloudEvent data, the behavior depends on the data content type of the CloudEvent
+    /// and the type of the <see cref="CloudEvent.Data"/> property value, following the rules below.
+    /// </para>
+    /// <list type="bullet">
+    /// <item><description>
+    /// If the data value is null, the content is empty for a binary mode message, and neither the "data"
+    /// nor "data_base64" property is populated in a structured mode message.
+    /// </description></item>
+    /// <item><description>
+    /// If the data content type is absent or has a media type of "application/json", the data is encoded as JSON.
+    /// If the data is already a <see cref="JToken"/>, that is serialized directly as JSON. Otherwise, the data
+    /// is converted using the <see cref="JsonSerializer"/> passed into the constructor, or a
+    /// default serializer.
+    /// </description></item>
+    /// <item><description>
+    /// Otherwise, if the data content type has a media type beginning with "text/" and the data value is a string,
+    /// the data is serialized as a string.
+    /// </description></item>
+    /// <item><description>
+    /// Otherwise, if the data value is a byte array, it is serialized either directly as binary data
+    /// (for binary mode messages) or as base64 data (for structured mode messages).
+    /// </description></item>
+    /// <item><description>
+    /// Otherwise, the encoding operation fails.
+    /// </description></item>
+    /// </list>
+    /// <para>
+    /// When decoding CloudEvent data, the following rules are used:
+    /// </para>
+    /// <para>
+    /// In a structured mode message, any data is either binary data within the "data_base64" property value,
+    /// or is a JSON token as the "data" property value. Binary data is represented as a byte array.
+    /// A JSON token is decoded as a string if is just a string value and the data content type is specified
+    /// and has a media type beginning with "text/". A JSON token representing the null value always
+    /// leads to a null data result. In any other situation, the JSON token is preserved as a <see cref="JToken"/>
+    /// that can be used for further deserialization (e.g. to a specific CLR type).
+    /// </para>
+    /// <para>
+    /// In a binary mode message, the data is parsed based on the content type of the message. When the content
+    /// type is absent or has a media type of "application/json", the data is parsed as JSON, with the result as
+    /// a <see cref="JToken"/>. When the content type has a media type beginning with "text/", the data is parsed
+    /// as a string. In all other cases, the data is left as a byte array.
+    /// </para>
+    /// </remarks>
     public class JsonEventFormatter : CloudEventFormatter
     {
+        private const string JsonMediaType = "application/json";
         private const string DataBase64 = "data_base64";
         private const string Data = "data";
-        public const string MediaTypeSuffix = "+json";
+        private const string MediaTypeSuffix = "+json";
 
-        public override async Task<CloudEvent> DecodeStructuredEventAsync(Stream data, IEnumerable<CloudEventAttribute> extensionAttributes)
+        private readonly JsonSerializer serializer;
+
+        /// <summary>
+        /// Creates a JsonEventFormatter that uses a default <see cref="JsonSerializer"/>.
+        /// </summary>
+        public JsonEventFormatter() : this(JsonSerializer.CreateDefault())
         {
-            var jsonReader = new JsonTextReader(new StreamReader(data, Encoding.UTF8, true, 8192, true))
-            {
-                DateParseHandling = DateParseHandling.None
-            };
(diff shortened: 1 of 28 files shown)


src/CloudNative.CloudEvents/CloudEventFormatter.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// Decodes a CloudEvent from a structured-mode message body, represented as a byte array.
+        /// </summary>
+        /// <param name="data">The data within the message body. Must not be null.</param>
+        /// <param name="contentType">The content type of the message, or null if no content type is known.
+        /// Typically this is a content type with a media type of "application/cloudevents"; the additional
+        /// information such as the charset parameter may be needed in order to decode the data.</param>
+        /// <param name="extensions">The extension attributes to use when populating the CloudEvent. May be null.</param>
+        /// <returns>The CloudEvent derived from the structured data.</returns>
+        public virtual CloudEvent DecodeStructuredModeMessage(byte[] data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            throw new NotImplementedException();
```
JoshLove-msft, line 83: Why aren't these implemented?

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -104,42 +152,77 @@ public CloudEvent DecodeJObject(JObject jObject, IEnumerable<CloudEventAttribute
...
-                    jObject[Data] = JToken.FromObject(cloudEvent.Data);
+                    jObject[Data] = cloudEvent.Data is JToken token
+                        ? token
+                        : JToken.FromObject(cloudEvent.Data);
                 }
-                else if (cloudEvent.Data is string text && cloudEvent.DataContentType?.StartsWith("text/") == true)
+                else if (cloudEvent.Data is string text && dataContentType.MediaType.StartsWith("text/"))
                 {
                     jObject[Data] = text;
                 }
```
JoshLove-msft, line 228: Should we be checking for `IEnumerable<ReadOnlyMemory<byte>>`? I think this would also catch byte array.

JoshLove-msft (approved):

src/CloudNative.CloudEvents/BinaryDataUtilities.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Utilities methods for dealing with binary data, converting between
+    /// streams, arrays, Memory{T} etc.
+    /// </summary>
+    internal static class BinaryDataUtilities
+    {
+        internal async static Task<byte[]> ToByteArrayAsync(Stream stream)
+        {
+            // TODO: Optimize if it's already a MemoryStream?
+            var memory = new MemoryStream();
```
JoshLove-msft, line 19: should there be any validation around the size of the stream if it is seekable?

src/CloudNative.CloudEvents/MimeUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+// Licensed under the Apache 2.0 license.
+// See LICENSE file in the project root for full license information.
+
+using System.Net.Http.Headers;
+using System.Net.Mime;
+using System.Text;
+
+namespace CloudNative.CloudEvents
+{
+    // TODO: Consider this name and namespace carefully. It really does need to be public, as all the event formatters are elsewhere.
```
JoshLove-msft, line 11: Could we use shared source instead?

src/CloudNative.CloudEvents/MimeUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+    public static class MimeUtilities
+    {
+        // TODO: Should we return null, and force the caller to do the appropriate defaulting?
+        /// <summary>
+        /// Returns an encoding from a content type, defaulting to UTF-8.
+        /// </summary>
+        /// <param name="contentType">The content type, or null if no content type is known.</param>
+        /// <returns>An encoding suitable for the charset specified in <paramref name="contentType"/>,
+        /// or UTF-8 if no charset has been specified.</returns>
+        public static Encoding GetEncoding(this ContentType contentType) =>
```
JoshLove-msft, line 26: UTF-8 as a default seems okay.

src/CloudNative.CloudEvents/CloudEventFormatter.cs
jskeet: Because they can't be - this code doesn't know how to decode messages :)
We could make the methods abstract instead... but if the point of having it as an abstract class rather than as an interface is to allow for further methods to be added without breaking binary compatibility, we should work out what we'd have done if we'd wanted to add these methods later. (Basically I'm looking for consistency - it's a little odd to have some methods abstract, and others just virtual. Maybe we can live with that, given that event formatters really should implement these...)

src/CloudNative.CloudEvents/BinaryDataUtilities.cs
jskeet: I don't think so. If someone chooses to pass in an enormous stream that will eat up all their memory, that's their call - I suspect any validation we'd put in place would have to be entirely arbitrary.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: I haven't added a System.Memory dependency yet - that should be the next topic of discussion for Friday. I wanted to try to get the overall responsibility shape sorted first, then we can work out whether every `byte[]` occurrence should actually be `ReadOnlyMemory<byte>`. (Not sure I follow about `IEnumerable<ReadOnlyMemory<byte>>` but we can talk about that.)

src/CloudNative.CloudEvents/MimeUtilities.cs
jskeet: I think it depends on the context. For most of *our* contexts, it's probably okay. I'd still like to leave the TODO there for now, but I'm kinda hoping we won't find anywhere that it's wrong :)

jskeet (commented):
Thanks Josh - will merge now and start on the next PR, but we should chat about the items you've raised in comments in our Friday meeting.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -14,85 +14,133 @@
...
+    /// nor "data_base64" property is populated in a structured mode message.
+    /// </description></item>
+    /// <item><description>
+    /// If the data content type is absent or has a media type of "application/json", the data is encoded as JSON.
+    /// If the data is already a <see cref="JToken"/>, that is serialized directly as JSON. Otherwise, the data
+    /// is converted using the <see cref="JsonSerializer"/> passed into the constructor, or a
+    /// default serializer.
+    /// </description></item>
+    /// <item><description>
+    /// Otherwise, if the data content type has a media type beginning with "text/" and the data value is a string,
```
JoshLove-msft, line 36: What about having an explicit parameter that can be set by the user to control how the data should be serialized?

src/CloudNative.CloudEvents/CloudEventFormatter.cs
JoshLove-msft: Ah yeah - it seems like it would be correct to make this abstract in that case.

jskeet: Okay - I may do that in a trivial follow-up PR before anything more meaningful then.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Not sure... at that point it feels like the user's doing the event formatter's job for it. They could always change the value to be a JToken and set the content type to application/json if they want it to be JSON. Let's talk more though - I'm not set on what we've got now.

PR #99, opened 2021-02-19
Make unimplemented methods abstract in CloudEventFormatter

jskeet:
Once we've hit 2.0, we won't be able to add new abstract methods to
this class, but for the moment it's okay... and it really does make
sense for these methods to be abstract.


JoshLove-msft (approved):

jskeet:
That was rather quicker approval than expected! Thanks :)

PR #100, opened 2021-02-23
Comprehensive tests for JsonEventFormatter

jskeet:
This required a bit of rework for the implementation, but the result
is cleaner, in my view.

There's one skipped test, for when an attribute (other than data) is
populated via an integer/Boolean token. We should discuss what we
want to happen in that case.

The extension tests had to change as JsonEventFormatter now
validates that the event it's parsed is valid (for structured mode),
and the ones we were using didn't have a source attribute.

(We will probably want to pepper lots of places with calls to
CloudEvent.Validate, but most of those can come later.)

### DIFF
--- a/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
@@ -13,6 +13,8 @@ using System.Threading.Tasks;
 
 namespace CloudNative.CloudEvents.NewtonsoftJson
 {
+    // TODO: Rename to JsonCloudEventFormatter? NewtonsoftJsonCloudEventFormatter?
+
     /// <summary>
     /// Formatter that implements the JSON Event Format.
     /// </summary>
@@ -58,7 +60,7 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
     /// <para>
     /// In a binary mode message, the data is parsed based on the content type of the message. When the content
     /// type is absent or has a media type of "application/json", the data is parsed as JSON, with the result as
-    /// a <see cref="JToken"/>. When the content type has a media type beginning with "text/", the data is parsed
+    /// a <see cref="JToken"/> (or null if the data is empty). When the content type has a media type beginning with "text/", the data is parsed
     /// as a string. In all other cases, the data is left as a byte array.
     /// </para>
     /// </remarks>
@@ -120,8 +122,7 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
             var cloudEvent = new CloudEvent(specVersion, extensionAttributes);
             PopulateAttributesFromStructuredEvent(cloudEvent, jObject);
             PopulateDataFromStructuredEvent(cloudEvent, jObject);
-
-            return cloudEvent;
+            return cloudEvent.Validate();
         }
 
         private void PopulateAttributesFromStructuredEvent(CloudEvent cloudEvent, JObject jObject)
@@ -141,7 +142,9 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
                     continue;
                 }
 
-                var attribute = cloudEvent.GetAttribute(key);
+                // TODO: Validate that if the attribute is known, the token type is appropriate. Possibly
+                // only for known-to-spec-version attributes? If an integer extension attribute has a value of "1",
+                // is that okay?
 
                 // Set the attribute in the event, taking account of mismatches between the type in the JObject
                 // and the attribute type as best we can.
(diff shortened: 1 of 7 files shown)


src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -13,6 +13,8 @@
 
 namespace CloudNative.CloudEvents.NewtonsoftJson
 {
+    // TODO: Rename to JsonCloudEventFormatter? NewtonsoftJsonCloudEventFormatter?
```
JoshLove-msft, line 16: Yeah we may want to differentiate in the type name if we think apps might want both packages.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -271,15 +312,24 @@ public override void DecodeBinaryModeEventData(byte[] value, CloudEvent cloudEve
...
+                }
             }
             else if (contentType.MediaType.StartsWith("text/") == true)
             {
                 cloudEvent.Data = encoding.GetString(value);
             }
             else
             {
+                // TODO: Should we clone the data? It's wasteful if the caller isn't reusing the array,
+                // but would isolate the data. Hmm.
```
JoshLove-msft, line 332: I don't think we should.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -141,7 +142,9 @@ private void PopulateAttributesFromStructuredEvent(CloudEvent cloudEvent, JObjec
                     continue;
                 }
 
-                var attribute = cloudEvent.GetAttribute(key);
+                // TODO: Validate that if the attribute is known, the token type is appropriate. Possibly
+                // only for known-to-spec-version attributes? If an integer extension attribute has a value of "1",
```
JoshLove-msft, line 146: Personally, I don't think that is necessary. If the CloudEvent originated from a different SDK, and you are registering an Integer extension, it feels like it should be enough for the value to be parsable into an int.

JoshLove-msft (approved):

test/CloudNative.CloudEvents.UnitTests/NewtonsoftJson/JsonEventFormatterTest.cs
```diff
@@ -2,76 +2,785 @@
...
+                ["specversion"] = 1,
+                ["type"] = "test-type",
+                ["id"] = "test-id",
+                ["source"] = SampleUriText,
+            };
+            Assert.Throws<ArgumentException>(() => DecodeStructuredModeMessage(obj));
+        }
+
+        [Fact(Skip = "Not yet implemented")]
+        public void DecodeStructuredModeMessage_TypeNotString()
```
JoshLove-msft, line 444: Although this somewhat contradicts my thoughts on the extension attribute topic, I think this should throw. It makes sense to be more strict for the reserved attributes. It might be worth thinking about having a parameter that can be set that allows validation to be skipped.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: I suspect that will be rare, to be honest - and apps can always use aliases etc. I'll leave the TODO there, but it's more something to revisit than "we definitely need to change this".

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Right. For extensions, I can see both sides - particularly given the way this SDK currently assumes "if we haven't seen something before, just store it as a string extension". There are cases where it *would* be nice to pick up issues early, but I think being lenient outweighs it.

For reserved attributes, as you've noted later, I think we should be stricter.

I won't change anything now, but I'll implement that in a follow-up PR and maybe note the deliberate decision for extension attributes.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Works for me. Will remove the TODO.

test/CloudNative.CloudEvents.UnitTests/NewtonsoftJson/JsonEventFormatterTest.cs
jskeet: Agreed for the first part. I think I'd want to wait until we get a feature request before allowing validation to be skipped, but I can see it *might* be useful.

Will implement in a follow-up PR (quite possibly tomorrow).

jskeet (commented):
Will remove the TODO around cloning, and merge.

Issue #91, opened 2021-02-04
Consider not using inheritance for Kafka and Amqp


jskeet:
Currently we use constructors to create messages for Kafka and Amqp, whereas other transports use extension methods.
I see no obvious benefits in using inheritance here, and it does make it harder to write asynchronous options.

jskeet:
Fixed in #101.

PR #101, opened 2021-02-26
Remove overuse of inheritance

jskeet:
None of these cases involved specialization - they were creating a new type for the sake of it.


jskeet:
Fixes #91.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -94,5 +95,96 @@ private static bool HasCloudEventsContentType(Message message, out string conten
...
+        private static RestrictedDescribed SerializeData(object data)
+        {
+            switch (data)
+            {
+                case null:
+                    return null;
+                case byte[] bytes:
+                    return new Data { Binary = bytes };
+                case MemoryStream memoryStream:
+                    // Note: this will return the whole stream, regardless of position...
```
JoshLove-msft, line 177: Hmm, I hadn't noticed this before. Why don't we want to respect the position?

src/CloudNative.CloudEvents.Kafka/KafkaClientExtensions.cs
```diff
@@ -100,5 +106,84 @@ private static void InitPartitioningKey(Message<string, byte[]> message, CloudEv
...
+                if (cloudEvent.Data is byte[] byteData)
+                {
+                    value = byteData;
+                }
+                else if (cloudEvent.Data is Stream dataStream)
+                {
+                    // TODO: Extract this common code somewhere, or use shared source to access BinaryDataUtilities.
+                    if (dataStream is MemoryStream dataMemoryStream)
+                    {
+                        value = dataMemoryStream.ToArray();
```
JoshLove-msft, line 139: Similar comment about whether we should take the Position into account.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -94,5 +95,96 @@ private static bool HasCloudEventsContentType(Message message, out string conten
...
+                    DateTimeOffset dto => dto.UtcDateTime,
+                    _ => pair.Value
+                };
+                properties.Add(propKey, propValue);
+            }
+            return applicationProperties;
+        }
+
+        /// <summary>
+        /// Convert data into a suitable format for inclusion in an Avro record.
```
JoshLove-msft, line 163: Looks like this was copied incorrectly as it references Avro.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -94,5 +95,96 @@ private static bool HasCloudEventsContentType(Message message, out string conten
...
+                    return new Data { Binary = bytes };
+                case MemoryStream memoryStream:
+                    // Note: this will return the whole stream, regardless of position...
+                    return new Data { Binary = memoryStream.ToArray() };
+                case Stream stream:
+                    var buffer = new MemoryStream();
+                    stream.CopyTo(buffer);
+                    return new Data { Binary = buffer.ToArray() };
+                case string text:
+                    return new AmqpValue { Value = text };
```
JoshLove-msft, line 184: The spec says the data should go in the Data section (as opposed to the value section which is what AmqpValue maps to) - https://github.com/cloudevents/spec/blob/v1.0/amqp-protocol-binding.md#13-content-modes

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -94,5 +95,96 @@ private static bool HasCloudEventsContentType(Message message, out string conten
...
+
+            switch (contentMode)
+            {
+                case ContentMode.Structured:
+                    bodySection = new Data
+                    {
+                        Binary = formatter.EncodeStructuredModeMessage(cloudEvent, out var contentType)
+                    };
+                    // TODO: What about the other parts of the content type?
+                    properties = new Properties { ContentType = contentType.MediaType };
```
JoshLove-msft, line 113: Hmm, the spec doesn't indicate that ContentType should be set for structured mode - https://github.com/cloudevents/spec/blob/v1.0/amqp-protocol-binding.md#13-content-modes

JoshLove-msft (approved):

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: It's specified here: https://github.com/cloudevents/spec/blob/v1.0/amqp-protocol-binding.md#321-amqp-content-type

(The doc isn't very clear, admittedly...)

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: It was copied accurately from the now-deleted AmqpCloudEventMessage - which of course was broken :)
Fixed now.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: Indeed. I'm pretty dubious about supporting `Stream` as a value in `CloudEvent.Data` at all, to be honest. It feels really fragile - who is responsible for closing the stream?

This PR doesn't change the current behavior though, it just moves it.

I've filed #103 to re-examine this.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: Well spotted. Filed https://github.com/cloudevents/sdk-csharp/issues/104 for this. (This PR doesn't change the behavior, just moves it. We'll want more tests in general...)

jskeet (commented):
Thanks for the review. Result:

- Filed two issues (about Stream and the position of text data)
- Fixed one comment
- The content-type part for AMQP is correct, I believe... but feel free to push back on that

PR #107, opened 2021-03-10
fix: Change AMQP data handling

### LINKED ISSUE #104: AMQP: text ends up in "value" instead of "data"
This code in AmqpClientExtensions:

```csharp
 case string text:
    return new AmqpValue { Value = text };
```

... means the data ends up in the "value" section instead of "data". See https://github.com/cloudevents/spec/blob/v1.0/amqp-protocol-binding.md#13-content-modes

jskeet:
Rather than the AMQP code "knowing" how to handle different data
types, delegate that to an event formatter.

There's *still* a little bit of an ugly
not-quite-single-responsibility in terms of event formatters;
CloudEvent formats tend to be specified in terms of structured mode
messages, not binary mode. It's possible that we should really
separate out "data formatting" from everything else... but it does
get intertwined in structured mode formats.

This change would fix #104, but we certainly want more testing and
documentation for this code in general.

### DIFF
--- a/src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
+++ b/src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
@@ -19,7 +19,7 @@ namespace CloudNative.CloudEvents.Amqp
 
         public static bool IsCloudEvent(this Message message) =>
             HasCloudEventsContentType(message, out _) ||
-            message.ApplicationProperties.Map.ContainsKey(SpecVersionAmqpHeader);        
+            message.ApplicationProperties.Map.ContainsKey(SpecVersionAmqpHeader);
 
         public static CloudEvent ToCloudEvent(this Message message,
             CloudEventFormatter formatter,
@@ -27,7 +27,7 @@ namespace CloudNative.CloudEvents.Amqp
         {
             if (HasCloudEventsContentType(message, out var contentType))
             {
-                return formatter.DecodeStructuredModeMessage(new MemoryStream((byte[])message.Body), new ContentType(contentType), extensionAttributes);
+                return formatter.DecodeStructuredModeMessage(new MemoryStream((byte[]) message.Body), new ContentType(contentType), extensionAttributes);
             }
             else
             {
@@ -44,10 +44,9 @@ namespace CloudNative.CloudEvents.Amqp
 
                 var cloudEvent = new CloudEvent(version, extensionAttributes)
                 {
-                    Data = message.Body,
                     DataContentType = message.Properties.ContentType
                 };
-                
+
                 foreach (var property in propertyMap)
                 {
                     if (!(property.Key is string key && key.StartsWith(AmqpHeaderPrefix)))
@@ -55,7 +54,7 @@ namespace CloudNative.CloudEvents.Amqp
                         continue;
                     }
                     string attributeName = key.Substring(AmqpHeaderPrefix.Length).ToLowerInvariant();
-                    
+
                     // We've already dealt with the spec version.
                     if (attributeName == CloudEventsSpecVersion.SpecVersionAttribute.Name)
                     {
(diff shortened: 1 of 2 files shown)


src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -113,7 +123,7 @@ public static Message ToAmqpMessage(this CloudEvent cloudEvent, ContentMode cont
                     properties = new Properties { ContentType = contentType.MediaType };
                     break;
                 case ContentMode.Binary:
-                    bodySection = SerializeData(cloudEvent.Data);
+                    bodySection = new Data { Binary = formatter.EncodeBinaryModeEventData(cloudEvent) };
```
JoshLove-msft, line 126: Should we validate that formatter is not null?

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -85,6 +84,17 @@ public static class AmqpClientExtensions
...
                 }
+                // Populate the data after the rest of the CloudEvent
+                if (message.BodySection is Data data)
+                {
+                    // TODO: Will data.Binary always be populated? Perhaps it'll be data.Buffer?
+                    formatter.DecodeBinaryModeEventData(data.Binary, cloudEvent);
+                }
+                else if (message.BodySection is object)
+                {
+                    throw new ArgumentException("Binary mode data in AMQP message must be in the application data section");
```
JoshLove-msft, line 95: Should we include relevant spec url?

JoshLove-msft, line 95: Isn't this true for structured mode as well?

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -19,15 +19,15 @@ public static class AmqpClientExtensions
...
+            message.ApplicationProperties.Map.ContainsKey(SpecVersionAmqpHeader);
 
         public static CloudEvent ToCloudEvent(this Message message,
             CloudEventFormatter formatter,
             params CloudEventAttribute[] extensionAttributes)
         {
             if (HasCloudEventsContentType(message, out var contentType))
             {
-                return formatter.DecodeStructuredModeMessage(new MemoryStream((byte[])message.Body), new ContentType(contentType), extensionAttributes);
+                return formatter.DecodeStructuredModeMessage(new MemoryStream((byte[]) message.Body), new ContentType(contentType), extensionAttributes);
```
JoshLove-msft, line 30: Should structured mode do the same application data check?

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -85,6 +84,17 @@ public static class AmqpClientExtensions
                         cloudEvent[attributeName] = property.Value;
                     }
                 }
+                // Populate the data after the rest of the CloudEvent
+                if (message.BodySection is Data data)
+                {
+                    // TODO: Will data.Binary always be populated? Perhaps it'll be data.Buffer?
+                    formatter.DecodeBinaryModeEventData(data.Binary, cloudEvent);
```
JoshLove-msft, line 91: Looks like it will - https://github.com/Azure/amqpnetlite/blob/master/src/Framing/Data.cs#L46-L50

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: Yes, probably. The cast at the moment is somewhat unpleasant. Will leave that for a follow-up PR.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: That's good to know, thanks. (I'll change the TODO to explain that.)

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: I don't think so, because it could easily move over time.

jskeet: Yes - we can probably refactor all of this a bit, to be honest.

src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
jskeet: This is something we need to do in all message converters, yes - along with calling `Validate` on the CloudEvents we receive and return. Will probably try to do that in a single sweep.

PR #108, opened 2021-03-12
Document and implement a convention for protocol bindings

jskeet:
The intention is for each commit to be nicely bite-sized.

(This addresses previous concerns about "what if we don't have an event formatter?")


src/CloudNative.CloudEvents.Amqp/AmqpClientExtensions.cs
```diff
@@ -21,26 +25,48 @@ public static class AmqpClientExtensions
...
                 {
                     throw new ArgumentException("Request is not a CloudEvent");
                 }
-                var version = CloudEventsSpecVersion.FromVersionId(versionIdText);
-                if (version is null)
-                {
-                    throw new ArgumentException($"Unsupported CloudEvents spec version '{versionIdText}'");
-                }
+
+                var version = CloudEventsSpecVersion.FromVersionId(versionId as string)
```
JoshLove-msft, line 68: Should we have a null check for versionId? Otherwise this would throw a NRE I believe.

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -384,6 +385,26 @@ public CloudEvent Validate()
...
+        /// <summary>
+        /// Validates that this CloudEvent is valid in the same way as <see cref="IsValid"/>,
+        /// but throwing an <see cref="ArgumentException"/> using the given parameter name
+        /// if the event is invalid. This is typically used within protocol bindings as the last
+        /// step in conversion.
+        /// </summary>
+        /// <param name="paramName">The parameter name to use in the exception if the event is invalid.</param>
+        /// <exception cref="ArgumentException">The event is invalid.</exception>
+        /// <returns>A reference to the same object, for simplicity of method chaining.</returns>
+        public CloudEvent ValidateForConversion(string paramName)
```
JoshLove-msft, line 397: Why public?

JoshLove-msft, line 397: This feels a bit odd as a public API. I understand that we don't want to duplicate this code for each protocol binding but I wonder if this would be a good candidate for using shared source instead.

jskeet: Maybe. Shared source feels like it ends up putting "official" packages (i.e. ones in this repo) on a different standing than other implementations, which can't use the same sort.

How about having this as an extension method within a namespace intended to help implementations for event formats and protocol bindings? (There are other things we could put there too.) That keeps it out of the face of regular users, but keeps a level playing field.

src/CloudNative.CloudEvents/Http/HttpContentExtensions.cs
```diff
@@ -16,11 +16,15 @@ public static class HttpContentExtensions
...
         /// Converts a CloudEvent to <see cref="HttpContent"/>.
         /// </summary>
-        /// <param name="cloudEvent">CloudEvent</param>
+        /// <param name="cloudEvent">The CloudEvent to convert. Must not be null, and must be a valid CloudEvent.</param>
         /// <param name="contentMode">Content mode. Structured or binary.</param>
-        /// <param name="formatter">Event formatter</param>
+        /// <param name="formatter">The formatter to use within the conversion. Must not be null.</param>
         public static HttpContent ToHttpContent(this CloudEvent cloudEvent, ContentMode contentMode, CloudEventFormatter formatter)
         {
+            Preconditions.CheckNotNull(cloudEvent, nameof(cloudEvent));
```
JoshLove-msft, line 24: Should Http bindings be in a separate package like the other protocol bindings?

jskeet: The other ones all have dependencies outside the framework. (That's why the ASP.NET Core one is separate.) I think it makes sense to bundle it where that doesn't introduce any dependencies. Happy to discuss though - I don't think it's a slam-dunk.

src/CloudNative.CloudEvents/CloudEvent.cs
JoshLove-msft: Even in a separate namespace, it would still appear to users when using Intellisense if it is in the same package. It would probably at least sort it to the bottom so it might be an improvement though. I don't feel strongly here, but just wanted to call out that it seemed unusual as a public API, so good with whatever you decide.

src/CloudNative.CloudEvents/Http/HttpContentExtensions.cs
JoshLove-msft: Yeah, I think it is a case of architectural purity vs convenience. If it is more convenient to leave in the main package, I am fine with it.

src/CloudNative.CloudEvents.Kafka/KafkaClientExtensions.cs
```diff
@@ -117,42 +145,21 @@ private static void InitPartitioningKey(Message<string, byte[]> message, CloudEv
...
-                {
-                    throw new InvalidOperationException($"{cloudEvent.Data.GetType()} type is not supported for Cloud Event's Value.");
-                }
-                if (cloudEvent.DataContentType is string dataContentType)
-                {
-                    contentTypeHeaderValue = dataContentType;                    
-                }
+                case ContentMode.Structured:
+                    value = formatter.EncodeStructuredModeMessage(cloudEvent, out var contentType);
+                    // TODO: What about the non-media type parts?
```
JoshLove-msft, line 154: would something like the charset get included in contentType.ToString()? Would it make sense to use that instead of MediaType?

test/CloudNative.CloudEvents.UnitTests/CloudEventsSpecVersionTest.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text;
+using Xunit;
+
+namespace CloudNative.CloudEvents.UnitTests
+{
+    public class CloudEventsSpecVersionTest
+    {
+        [Theory]
+        [InlineData(null)]
+        [InlineData("bogus")]
```
JoshLove-msft, line 16: May want to throw in "1" for good measure.

JoshLove-msft (approved):

src/CloudNative.CloudEvents.Kafka/KafkaClientExtensions.cs
jskeet: Yes, I suspect that's the way forward. I've just been reluctant to do it so far, just in case it was actually deliberate. (It's in a bunch of places - I think we'll probably want to fix them all in one pass.)

test/CloudNative.CloudEvents.UnitTests/CloudEventsSpecVersionTest.cs
jskeet: Done.

jskeet:
Will leave the ValidateForConversion method in place at the moment, but with a TODO... once I've done the equivalent PR for event formats, we should be in a better position to work out what we want to do.

PR #109, opened 2021-03-17
Add guidance for CloudEventFormatter implementers

jskeet:
(And adjust the implementations to match.)

### DIFF
--- a/docs/formatters.md
+++ b/docs/formatters.md
@@ -1,0 +1,82 @@
+# Implementing an event formatter
+
+The `CloudEventFormatter` abstract type in the C# SDK is an
+augmentation of the [Event
+Format](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#event-format)
+concept in the specification.
+
+Strictly speaking, CloudEvent data is simply a sequence of bytes. In
+practical terms, it's useful to be able to store any object
+reference in the `CloudEvent.Data` property, leaving a
+`CloudEventFormatter` to perform serialization and deserialization
+when requested.
+
+This means that `CloudEventFormatter` implementations need to be
+aware of all content modes (binary, structured and batch) and
+document how they handle data of various types. A
+`CloudEventFormatter` implementation *may* implement only a subset
+of content modes, but should document this very clearly. (Note:
+batch content mode is not currently implemented in the SDK.)
+
+## Data serialization and deserialization
+
+When serializing data in binary mode messages, all formatters should
+handle data provided as a `byte[]`, serializing it without any
+modification. Formatters are also encouraged to support serializing
+strings in the obvious way (obeying any character encoding indicated
+in the `datacontenttype` attribute).
+
+When deserializing data in binary mode messages, event formatters
+may use the content type to determine the in-memory object type to
+deserialize to. For example, a JSON formatter may decode data in a
+message with a content type of "application/json" to a JSON API type
+(such as `Newtonsoft.Json.Linq.JToken`). Formatters are encouraged
+to deserialize data with a content type beginning `text/` (and which
+don't otherwise have a special meaning to the formatter) as
+strings, obeying any character encoding indicated in the content
+type. If the content type is unknown to the formatter, the data
+should be populated in the `CloudEvent` as a simple byte array.
+
+When serializing and deserializing data in a structured mode
+message, an event formatter should follow the rules of the event
+format it is implementing. The event formatter should be as
+consistent as is reasonably possible in terms of its handling of
+binary mode data and structured mode data, however. In particular, a
+well-designed event format should usually not be restricted to any
+specific data type, so any data that can be serialized in a binary
+mode message should be serializable in a structured mode message
+too.
+
+Inconsistencies may still arise, when the structured message
+contains more information about the original data than the
+corresponding binary message. For example, an event format may use a
+different serialization format for text and binary data, allowing
+string and byte arrays to be serialized and then deserialized
+seamlessly even if the content type is unknown to the formatter.
+However, a binary mode messages serialized from the same data string may
+lose that distinction, resulting in a `Data` property with a byte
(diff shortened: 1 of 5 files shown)


jskeet:
(Once we're happy with these, I'll implement a System.Text.Json-based event formatter, and the proto event format... which will probably involve some refactoring.)

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -355,7 +361,9 @@ public override byte[] EncodeStructuredModeMessage(CloudEvent cloudEvent, out Co
         /// <see cref="DataBase64PropertyName"/> properties.
         /// </para>
         /// </remarks>
-        /// <param name="cloudEvent">The CloudEvent being encoded, which will have a non-null value of
+        /// <param name="cloudEvent">The CloudEvent being encoded, which will have a non-null value for
+        /// its <see cref="CloudEvent.Data"/> property.
+        /// <paramref name="writer"/>The write to serialize the data to. Will not be null.</param>
```
JoshLove-msft, line 366:
```suggestion
        /// <paramref name="writer"/>The writer to serialize the data to. Will not be null.</param>
```

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -141,16 +145,15 @@ private CloudEvent DecodeJObject(JObject jObject, IEnumerable<CloudEventAttribut
...
-            }
+            var specVersion = CloudEventsSpecVersion.FromVersionId((string) specVersionToken)
+                ?? throw new ArgumentException($"Unsupported CloudEvents spec version '{(string)specVersionToken}'");
 
             var cloudEvent = new CloudEvent(specVersion, extensionAttributes);
             PopulateAttributesFromStructuredEvent(cloudEvent, jObject);
             PopulateDataFromStructuredEvent(cloudEvent, jObject);
-            return cloudEvent.Validate();
+            // "data" is always the parameter from the public method. It's annoying not to be able to use
+            // nameof here, but this will give the appropriate result.
```
JoshLove-msft, line 155: Sorry, but why can't nameof be used? The parameter in this method is `data`.

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -390,10 +390,11 @@ public CloudEvent Validate()
         /// <summary>
         /// Validates that this CloudEvent is valid in the same way as <see cref="IsValid"/>,
         /// but throwing an <see cref="ArgumentException"/> using the given parameter name
-        /// if the event is invalid. This is typically used within protocol bindings as the last
-        /// step in conversion.
+        /// if the event is invalid. This is typically used within protocol bindings or event formatters
+        /// as the last step in decoding an event, or as the first step when encoding an event.
         /// </summary>
-        /// <param name="paramName">The parameter name to use in the exception if the event is invalid.</param>
+        /// <param name="paramName">The parameter name to use in the exception if the event is invalid.
```
JoshLove-msft, line 396: Why is the parameter name important? Doesn't the error message include enough information since it lists the missing required attributes?

docs/formatters.md
```diff
@@ -0,0 +1,79 @@
...
+
+Event formatters should document their behavior clearly. While this
+doesn't allow `CloudEventFormatter` instances to be used
+interchangably, it at least provides consumers with some certainty
+around what they can expect for a specific formatter.
+
+## Validation
+
+Formatter implementations should validate references documented as
+being non-null, and additionally perform CloudEvent validation on:
```
JoshLove-msft, line 69: Should we mention that ValidateForConversion needs to be called?

src/CloudNative.CloudEvents/CloudEvent.cs
```diff
@@ -390,10 +390,11 @@ public CloudEvent Validate()
...
+        /// May be null.</param>
         /// <exception cref="ArgumentException">The event is invalid.</exception>
         /// <returns>A reference to the same object, for simplicity of method chaining.</returns>
         public CloudEvent ValidateForConversion(string paramName)
```
JoshLove-msft, line 400: What if this was just called `Validate`?

JoshLove-msft (approved):

docs/formatters.md
jskeet: Yup, done thanks. (If we move it to a different place or rename it, we'll need to change this and the protocol binding docs.)

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Unfortunately not - GitHub was hiding the fact that this code is in the `DecodeJObject` method :(

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Nice catch, thanks.

src/CloudNative.CloudEvents/CloudEvent.cs
jskeet: I just generally think it's a good idea to provide a parameter name in `ArgumentException`, so callers know which part of their call is invalid.

We *could* potentially drop that... it's only a nice-to-have.

src/CloudNative.CloudEvents/CloudEvent.cs
jskeet: There's already a more "normal" `Validate` method which throws `InvalidOperationException`.

I think moving `ValidateForConversion` to an extension method in a different namespace will be the best option. Will do that in a later PR.

PR #110, opened 2021-03-19
Implement JsonEventFormatter for System.Text.Json

### LINKED ISSUE #42: Serialize json using System.Text.Json

jskeet:
This leads to two different formatters both called
JsonEventFormatter, just in different namespaces. That probably
wouldn't cause too much ambiguity, as few applications would depend
on both packages. It avoids longer names such as
"SystemTextJsonEventFormatter" which is pretty unwieldy.

Both the implementation and the tests are very much based on the
Newtonsoft.Json implementation.

Fixes #42.

### DIFF
--- a/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
@@ -108,51 +108,39 @@ namespace CloudNative.CloudEvents.SystemTextJson
             DocumentOptions = documentOptions;
         }
 
-        public override async Task<CloudEvent> DecodeStructuredModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
-        {
-            data = data ?? throw new ArgumentNullException(nameof(data));
-            var encoding = contentType.GetEncoding();
+        public override Task<CloudEvent> DecodeStructuredModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeStructuredModeMessageImpl(data, contentType, extensionAttributes, true);
 
-            JsonDocument document;
-            if (encoding is UTF8Encoding)
-            {
-                document = await JsonDocument.ParseAsync(data, DocumentOptions).ConfigureAwait(false);
-            }
-            else
-            {
-                using var reader = new StreamReader(data, encoding);
-                string json = await reader.ReadToEndAsync().ConfigureAwait(false);
-                document = JsonDocument.Parse(json, DocumentOptions);
-            }
-            using (document)
-            {
-                return DecodeJsonDocument(document, extensionAttributes);
-            }
-        }
-
-        public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
-        {
-            data = data ?? throw new ArgumentNullException(nameof(data));
-            var encoding = contentType.GetEncoding();
-            JsonDocument document;
-            if (encoding is UTF8Encoding)
-            {
-                document = JsonDocument.Parse(data, DocumentOptions);
-            }
-            else
-            {
-                using var reader = new StreamReader(data, encoding);
-                string json = reader.ReadToEnd();
-                document = JsonDocument.Parse(json, DocumentOptions);
-            }
-            using (document)
-            {
-                return DecodeJsonDocument(document, extensionAttributes);
-            }
-        }
+        public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeStructuredModeMessageImpl(data, contentType, extensionAttributes, false).GetAwaiter().GetResult();
 
         public override CloudEvent DecodeStructuredModeMessage(byte[] data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
-            DecodeStructuredModeMessage(new MemoryStream(data), contentType, extensionAttributes);
+            DecodeStructuredModeMessageImpl(new MemoryStream(data), contentType, extensionAttributes, false).GetAwaiter().GetResult();
+
+        private async Task<CloudEvent> DecodeStructuredModeMessageImpl(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes, bool async)
+        {
+            data = data ?? throw new ArgumentNullException(nameof(data));
+            var encoding = contentType.GetEncoding();
+            JsonDocument document;
+            if (encoding is UTF8Encoding)
+            {
+                document = async
+                    ? await JsonDocument.ParseAsync(data, DocumentOptions).ConfigureAwait(false)
+                    : JsonDocument.Parse(data, DocumentOptions);
+            }
+            else
+            {
+                using var reader = new StreamReader(data, encoding);
+                string json = async
+                    ? await reader.ReadToEndAsync().ConfigureAwait(false)
+                    : reader.ReadToEnd();
+                document = JsonDocument.Parse(json, DocumentOptions);
+            }
+            using (document)
+            {
+                return DecodeJsonDocument(document, extensionAttributes);
+            }
+        }
 
         private CloudEvent DecodeJsonDocument(JsonDocument document, IEnumerable<CloudEventAttribute> extensionAttributes = null)
         {
--- a/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
@@ -2,26 +2,25 @@
 // Licensed under the Apache 2.0 license.
 // See LICENSE file in the project root for full license information.
 
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Mime;
 using System.Text;
+using System.Text.Json;
 using System.Threading.Tasks;
 
-namespace CloudNative.CloudEvents.NewtonsoftJson
+namespace CloudNative.CloudEvents.SystemTextJson
 {
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -0,0 +1,489 @@
...
+                string json = await reader.ReadToEndAsync().ConfigureAwait(false);
+                document = JsonDocument.Parse(json, DocumentOptions);
+            }
+            using (document)
+            {
+                return DecodeJsonDocument(document, extensionAttributes);
+            }
+        }
+
+        public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
```
JoshLove-msft, line 133: nit - consider calling into the async method and passing a `bool async` param that can be used to check whether to call sync or async methods.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -0,0 +1,489 @@
...
+            writer.WriteStartObject();
+            writer.WritePropertyName(CloudEventsSpecVersion.SpecVersionAttribute.Name);
+            writer.WriteStringValue(cloudEvent.SpecVersion.VersionId);
+            var attributes = cloudEvent.GetPopulatedAttributes();
+            foreach (var keyValuePair in attributes)
+            {
+                var attribute = keyValuePair.Key;
+                var value = keyValuePair.Value;
+                writer.WritePropertyName(attribute.Name);
+                // TODO: Maybe we should have an enum associated with CloudEventsAttributeType?
```
JoshLove-msft, line 360: is this TODO still relevant?

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -0,0 +1,489 @@
...
+            }
+        }
+
+        private void PopulateDataFromStructuredEvent(CloudEvent cloudEvent, JsonDocument document)
+        {
+            // Fetch data and data_base64 tokens, and treat null as missing.
+            document.RootElement.TryGetProperty(DataPropertyName, out var dataElement);
+            if (dataElement is JsonElement { ValueKind: JsonValueKind.Null })
+            {
+                dataElement = new JsonElement();
```
JoshLove-msft, line 257: Hmm, why do we create a new element if it is a Null value kind?

JoshLove-msft (approved):

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: So that it's equivalent to not being present (i.e. as if TryGetProperty had returned false). (In Json.NET we just use a null reference, but JsonElement is a struct.)

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: I've done this in a separate commit - see whether it's what you were expecting. It looks nice other than the exception unwrapping part... if you're aware of a framework method that does a sort of "give me the task result, unwrapping any exceptions appropriately" that would make it more of a slam dunk change.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: Yes - it would be nice to be able to switch on the attribute type, here and in ValidateTokenTypeForAttribute. I've got a separate branch which does that already... I'll create a separate PR to see what you think of it. (I'm not entirely sure yet, myself...)

jskeet:
(Not merging, as I've pushed a second commit which I'd like feedback on.)

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: Doh - `GetAwaiter().GetResult()` is fine. PTAL - if this is what you expected, I'll rebase it into the previous commit, and merge.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -108,41 +108,32 @@ public JsonEventFormatter(JsonSerializerOptions serializerOptions, JsonDocumentO
...
-                using var reader = new StreamReader(data, encoding);
-                string json = await reader.ReadToEndAsync().ConfigureAwait(false);
-                document = JsonDocument.Parse(json, DocumentOptions);
-            }
-            using (document)
-            {
-                return DecodeJsonDocument(document, extensionAttributes);
-            }
-        }
+        public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
```
JoshLove-msft, line 114: Should we await here and make this async?

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -108,41 +108,32 @@ public JsonEventFormatter(JsonSerializerOptions serializerOptions, JsonDocumentO
             DocumentOptions = documentOptions;
         }
 
-        public override async Task<CloudEvent> DecodeStructuredModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
-        {
-            data = data ?? throw new ArgumentNullException(nameof(data));
-            var encoding = contentType.GetEncoding();
+        public override Task<CloudEvent> DecodeStructuredModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeStructuredModeMessageImpl(data, contentType, extensionAttributes, true);
```
JoshLove-msft, line 112: Should we await this call and mark the method as async?

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
JoshLove-msft: interesting.. I wouldn't have expected that we would mutate the document in this method.

JoshLove-msft (approved):

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: No need - the task returned by the "impl" method is fine already. (If that method had been implemented without using `async`, there'd be some potential for difference in terms of timing for exceptions etc, but as it is, they'll behave the same way.)

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: We're not - we're just changing the local variable to have a different value. That doesn't mutate the document at all. It's just saying "treat a null value as if it weren't present at all".

jskeet (commented):
Thanks for the review - will merge now. I think I'll actually leave it as the two commits, given that they do make sense separately.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
JoshLove-msft: But why do that? it doesn't appear to be used other than in doing the Undefined check.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
JoshLove-msft: I believe the best practice is to always await the task in a public API to provide an understandable call stack.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: If we did that, we'd basically need the opposite code, to coalesce undefined to null. An alternative would be to extract two `bool` variables, e.g.

```csharp
bool gotData = dataElement.ValueKind != JsonValueKind.Null && dataElement.ValueKind != JsonValueKind.Undefined;
// Ditto for base64
```

Would you prefer that? Happy to do so.

(I note that the pattern matching isn't required here - that's left over from Json.NET.)

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: PR with that in is https://github.com/cloudevents/sdk-csharp/pull/112

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
JoshLove-msft: I see - we need to handle both Undefined and Null. That was the part I was missing. Thanks!

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: Hmm... happy to do that if it's really best practice (not one I've come across before for this situation), but I don't *think* the stack trace will be hard to understand here - it feels like it would just be adding a perf hit.

Do you have any links to docs about this? Always eager to learn more about best practices. If there's something that's MSFT internal, feel free to say so and I'll take it on trust :)

JoshLove-msft: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task

jskeet: Cool, thanks. Will implement tomorrow - and look at ways of simplifying the data checks as well. (The fact that it wasn't immediately clear to you is enough signal that it could do with work.)

PR #111, opened 2021-03-23
Event formatter batch handling

jskeet:
(This comment is now out of date; see comment from 2021-04-08 for update.)

First two commits are from #110.

Obviously we'd need tests, and then probably protocol binding support (at least for HTTP). This is mostly to show what the CloudEventFormatter interface might look like for batch.

I'm nervous about it because it's not a separate content mode in the CloudEvents spec - but it would be really hard to express without putting it into CloudEventFormatter. It probably doesn't make sense to include it in the ContentMode enum, because the signatures for protocol bindings don't make sense (as they need to accept/return a collection instead of a single event). Basically it doesn't fit in terribly neatly...

### DIFF
--- a/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
@@ -85,6 +85,9 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
         private const string JsonMediaType = "application/json";
         private const string MediaTypeSuffix = "+json";
 
+        private static readonly string StructuredMediaType = CloudEvent.MediaType + MediaTypeSuffix;
+        private static readonly string BatchMediaType = MimeUtilities.BatchMediaType + MediaTypeSuffix;
+
         /// <summary>
         /// The property name to use for base64-encoded binary data in a structured-mode message.
         /// </summary>
@@ -136,6 +139,44 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
 
         public override CloudEvent DecodeStructuredModeMessage(byte[] data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
             DecodeStructuredModeMessage(new MemoryStream(data), contentType, extensionAttributes);
+
+        public override async Task<IReadOnlyList<CloudEvent>> DecodeBatchModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
+        {
+            Validation.CheckNotNull(data, nameof(data));
+
+            var jsonReader = CreateJsonReader(data, MimeUtilities.GetEncoding(contentType));
+            var array = await JArray.LoadAsync(jsonReader).ConfigureAwait(false);
+            return DecodeJArray(array, extensionAttributes, nameof(data));
+        }
+
+        public override IReadOnlyList<CloudEvent> DecodeBatchModeMessage(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
+        {
+            Validation.CheckNotNull(data, nameof(data));
+
+            var jsonReader = CreateJsonReader(data, MimeUtilities.GetEncoding(contentType));
+            var array = JArray.Load(jsonReader);
+            return DecodeJArray(array, extensionAttributes, nameof(data));
+        }
+
+        public override IReadOnlyList<CloudEvent> DecodeBatchModeMessage(byte[] data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeBatchModeMessage(new MemoryStream(data), contentType, extensionAttributes);
+
+        private IReadOnlyList<CloudEvent> DecodeJArray(JArray jArray, IEnumerable<CloudEventAttribute> extensionAttributes, string paramName)
+        {
+            List<CloudEvent> events = new List<CloudEvent>(jArray.Count);
+            foreach (var token in jArray)
+            {
+                if (token is JObject obj)
+                {
+                    events.Add(DecodeJObject(obj, extensionAttributes));
+                }
+                else
+                {
+                    throw new ArgumentException($"Invalid array element index {events.Count} within batch; expected an object, but token type was '{token?.Type}'", paramName);
+                }
+            }
+            return events;
+        }
 
         private CloudEvent DecodeJObject(JObject jObject, IEnumerable<CloudEventAttribute> extensionAttributes = null)
         {
@@ -307,15 +348,45 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
 
         public override byte[] EncodeStructuredModeMessage(CloudEvent cloudEvent, out ContentType contentType)
         {
-            Validation.CheckCloudEventArgument(cloudEvent, nameof(cloudEvent));
+            // The cloudEvent parameter will be validated in WriteCloudEventForBatchOrStructuredMode
 
-            contentType = new ContentType("application/cloudevents+json")
+            contentType = new ContentType(StructuredMediaType)
             {
                 CharSet = Encoding.UTF8.WebName
             };
 
             var stream = new MemoryStream();
             var writer = new JsonTextWriter(new StreamWriter(stream));
+            WriteCloudEventForBatchOrStructuredMode(writer, cloudEvent);
+            writer.Flush();
+            return stream.ToArray();
+        }
+
+        public override byte[] EncodeBatchModeMessage(IEnumerable<CloudEvent> cloudEvents, out ContentType contentType)
+        {
+            Validation.CheckNotNull(cloudEvents, nameof(cloudEvents));
+
+            contentType = new ContentType(BatchMediaType)
+            {
+                CharSet = Encoding.UTF8.WebName
+            };
+
+            var stream = new MemoryStream();
+            var writer = new JsonTextWriter(new StreamWriter(stream));
+            writer.WriteStartArray();
+            foreach (var cloudEvent in cloudEvents)
+            {
+                WriteCloudEventForBatchOrStructuredMode(writer, cloudEvent);
+            }
+            writer.WriteEndArray();
+            writer.Flush();
+            return stream.ToArray();
+        }
+
+        private void WriteCloudEventForBatchOrStructuredMode(JsonWriter writer, CloudEvent cloudEvent)
+        {
+            Validation.CheckCloudEventArgument(cloudEvent, nameof(cloudEvent));
+
             writer.WriteStartObject();
             writer.WritePropertyName(CloudEventsSpecVersion.SpecVersionAttribute.Name);
             writer.WriteValue(cloudEvent.SpecVersion.VersionId);
@@ -344,8 +415,6 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
                 EncodeStructuredModeData(cloudEvent, writer);
             }
             writer.WriteEndObject();
-            writer.Flush();
-            return stream.ToArray();
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -120,31 +120,72 @@ public JsonEventFormatter(JsonSerializerOptions serializerOptions, JsonDocumentO
...
             }
         }
 
-        private CloudEvent DecodeJsonDocument(JsonDocument document, IEnumerable<CloudEventAttribute> extensionAttributes = null)
+        // TODO: Override the other methods
+
+        private CloudEvent DecodeJsonElement(JsonElement element, IEnumerable<CloudEventAttribute> extensionAttributes = null)
         {
-            if (!document.RootElement.TryGetProperty(CloudEventsSpecVersion.SpecVersionAttribute.Name, out var specVersionProperty)
+            if (element.ValueKind != JsonValueKind.Object)
```
JoshLove-msft, line 183: What about a single element array?

JoshLove-msft (approved):

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: I personally think trying to parse a single-element batch as if it's not a batch at all should still fail. The calling code is doing the wrong thing, and that should be reported as early as possible - otherwise I can see code that works in all tests and then fails with real batches. Happy to be persuaded though.

jskeet:
Will try to turn this into a non-prototype later this week.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
JoshLove-msft: I was thinking of the case where a caller might know that they will always get a single event in a JSON array. For instance, maybe the publisher always publishes it this way. But I think you raise a fair point about the potential for missing this during testing.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -136,6 +136,44 @@ public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType
...
+        public override async Task<IReadOnlyList<CloudEvent>> DecodeBatchModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
+        {
+            data = data ?? throw new ArgumentNullException(nameof(data));
+
+            var jsonReader = CreateJsonReader(data, contentType.GetEncoding());
+            var array = await JArray.LoadAsync(jsonReader).ConfigureAwait(false);
+            return DecodeJArray(array, extensionAttributes);
+        }
+
+        public override IReadOnlyList<CloudEvent> DecodeBatchModeMessage(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
```
JoshLove-msft, line 148: Do we know if there will ever be a Binary batch message as part of the spec?

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -136,6 +136,44 @@ public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType
         public override CloudEvent DecodeStructuredModeMessage(byte[] data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
             DecodeStructuredModeMessage(new MemoryStream(data), contentType, extensionAttributes);
```
JoshLove-msft, line 138: Update to mention new method.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -136,6 +136,44 @@ public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType
...
 
+        public override async Task<IReadOnlyList<CloudEvent>> DecodeBatchModeMessageAsync(Stream data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
```
JoshLove-msft, line 139: Consider updating to mention new methods https://github.com/cloudevents/sdk-csharp/pull/111/files#diff-b73904f7e0560631ddf370286e4d19cd3edf58de51d76bc28a6d33b11edb5b94R59

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -136,6 +136,44 @@ public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType
...
+            var array = JArray.Load(jsonReader);
+            return DecodeJArray(array, extensionAttributes);
+        }
+
+        public override IReadOnlyList<CloudEvent> DecodeBatchModeMessage(byte[] data, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeBatchModeMessage(new MemoryStream(data), contentType, extensionAttributes);
+
+        private IReadOnlyList<CloudEvent> DecodeJArray(JArray jArray, IEnumerable<CloudEventAttribute> extensionAttributes = null)
+        {
+            List<CloudEvent> events = new List<CloudEvent>();
```
JoshLove-msft, line 162: Should we use jArray.Count when initializing the list?

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -136,6 +136,44 @@ public override CloudEvent DecodeStructuredModeMessage(Stream data, ContentType
...
+            foreach (var token in jArray)
+            {
+                if (token is JObject obj)
+                {
+                    events.Add(DecodeJObject(obj, extensionAttributes));
+                }
+                else
+                {
+                    throw new ArgumentException("Invalid array element within batch", "data");
```
JoshLove-msft, line 171: consider passing in paramName

JoshLove-msft: Should we include the token in the exception?

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -306,16 +344,43 @@ protected virtual void DecodeStructuredModeDataBase64Property(JToken dataBase64T
...
             var writer = new JsonTextWriter(new StreamWriter(stream));
+            WriteCloudEventForBatchOrStructuredMode(writer, cloudEvent);
+            writer.Flush();
+            return stream.ToArray();
+        }
+
+        public override byte[] EncodeBatchModeMessage(IEnumerable<CloudEvent> cloudEvents, out ContentType contentType)
+        {
+            cloudEvents = cloudEvents ?? throw new ArgumentNullException(nameof(cloudEvents));
+            contentType = new ContentType("application/cloudevents-batch+json")
```
JoshLove-msft, line 362: Consider putting this in a constants file along with the content type for structured mode.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
```diff
@@ -120,31 +120,72 @@ public JsonEventFormatter(JsonSerializerOptions serializerOptions, JsonDocumentO
...
+            {
+                throw new ArgumentException($"Cannot decode JSON element of kind '{element.ValueKind}' as CloudEvent");
```
JoshLove-msft, line 185: Might be worth including the element itself as well.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Yes indeed! Thanks for the reminder. Will definitely do that.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: I've seen no hint of that, no. I can't see how it would work to be honest, given that headers are used for the attributes, and each event would need its own set of attributes. (I'm sure it could be done, but I suspect it would be massively clunky compared with just using the current approach.) I think we're okay to design the SDK (and method names in particular) assuming that batch always means "in a structured mode style".

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Happy to pass in paramName to avoid the hard-coding, yes - thanks.
I don't want to include the actual token, as generally I don't like exceptions containing potentially arbitrary data (both in terms of size and PII)... but we could include the type.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: It certainly wouldn't hurt. (Alternatively this method could be rewritten via LINQ... I might have a try at that, as it feels odd to use a loop here.)

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Definitely happy to make it a constant... not sure about a file just for constants though. I could put the prefix in MimeUtilities which is where most of our content type common code is. Will have a look.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: Yup - will leave it for now. I suspect we may want more introspection capabilities on event formatters before we go GA, so you could have multiple formatters and pick the right one to use (and the right method to use, potentially) based on the content type.

src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
jskeet: As per earlier comment, I'd generally prefer to avoid including arbitrary data in the exception message. I could be persuaded though - let's chat about it.

jskeet (commented):
Thanks for the comments. I'll *try* to come back to this PR before the end of the week - adding tests as well as implementing your comments. I may not get to it though.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Done - decided not to use LINQ for the same reasons as in the System.Text.Json code; it seems unnecessarily inefficient and wouldn't actually be much simpler.

jskeet:
I've now rebased, addressed review comments, and added a commit with tests.

@JoshLove-msft please re-review with those changes... if you're happy, I'll merge and we can then get to the protocol binding methods.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
JoshLove-msft: Right.. I forgot about that regarding including arbitrary data. We can also include the index of the failing element.

jskeet: Done.

JoshLove-msft (approved):

jskeet:
(Rebased back to two commits, ready to merge on green.)

PR #112, opened 2021-03-24
DO NOT MERGE: Expose an Ordinal enum for CloudEventAttributeType

jskeet:
This is nested as most code won't need it, but it allows code to
use switch statements to react to different attribute types.

I'm interested in the following aspects for review:

- Whether we think this is sufficiently useful to be worth it at all
- The design of using a nested enum

(Another option would be to have an enum in a namespace which is
designed primarily for protocol/formatter implementers - we've
already discussed the possibility of having that namespace, and
maybe this could belong there?)

### DIFF
--- a/src/CloudNative.CloudEvents/CloudEventAttributeType.cs
+++ b/src/CloudNative.CloudEvents/CloudEventAttributeType.cs
@@ -52,6 +52,52 @@ namespace CloudNative.CloudEvents
         public static CloudEventAttributeType Timestamp { get; } = new TimestampType();
 
         /// <summary>
+        /// Enum of event types, to allow efficient switching over CloudEventAttributeType.
+        /// Each attribute type has a unique value, returned by <see cref="OrdinalType"/>.
+        /// </summary>
+        /// <remarks>
+        /// This type is nested as relatively few consumers will need to use it.
+        /// </remarks>
+        public enum Ordinal
+        {
+            // Note: changing the values here is a breaking change.
+
+            /// <summary>
+            /// Ordinal for <see cref="CloudEventAttributeType.Binary"/>
+            /// </summary>
+            Binary = 0,
+            /// <summary>
+            /// Ordinal for <see cref="CloudEventAttributeType.Boolean"/>
+            /// </summary>
+            Boolean = 1,
+            /// <summary>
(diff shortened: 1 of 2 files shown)


src/CloudNative.CloudEvents/CloudEventAttributeType.cs
```diff
@@ -86,15 +132,16 @@ public abstract class CloudEventAttributeType
         /// <param name="value">The value to validate. Must be non-null, and suitable for this attribute type.</param>
         public abstract void Validate(object value);
 
-        private CloudEventAttributeType(string name, Type clrType)
+        private CloudEventAttributeType(string name, Ordinal ordinal, Type clrType)
         {
             Name = name;
```
JoshLove-msft, line 137: Can we do ordinal.ToString() for the name?

jskeet: Unfortunately not - the names include `"URI-Reference"` which wouldn't be a valid enum value. We could potentially have an attribute on the enum value to specify the name... but I suspect just hard-coding the names is simpler.

src/CloudNative.CloudEvents/CloudEventAttributeType.cs
```diff
@@ -51,6 +51,52 @@ public abstract class CloudEventAttributeType
...
         public static CloudEventAttributeType Timestamp { get; } = new TimestampType();
 
+        /// <summary>
+        /// Enum of event types, to allow efficient switching over CloudEventAttributeType.
+        /// Each attribute type has a unique value, returned by <see cref="OrdinalType"/>.
+        /// </summary>
+        /// <remarks>
+        /// This type is nested as relatively few consumers will need to use it.
+        /// </remarks>
+        public enum Ordinal
```
JoshLove-msft, line 61: Agree with the idea of moving this to the Core namespace rather than nesting it here.

jskeet: Yup - I've got another PR *nearly* ready for this... and which avoids having a public property on the class as well. (It's an internal property which you get at via a static method in a static class in the Core namespace - basically that extra layer of indirection effectively obscures it from "regular" users.)

jskeet:
Closing in favour of #112 (now merged).

PR #114, opened 2021-03-25
Move implementation utility methods into their own namespace

jskeet:
- Preconditions is renamed to Validation
- Validation is public, allowing simplified validation in
  implementation classes
- CloudEvent.ValidateForConversion is now Validation.CheckCloudEventArgument
- All the MimeUtilities methods are non-extension methods (to avoid
  Intellisense suggesting them)
- BinaryDataUtilities is now public (it's much less of an issue once
  it's in a namespace that other users are somewhat discouraged from
  using)
- Documentation explains the purpose of the namespace, albeit briefly

Other suggestions for a namespace would be welcome. The intention is
to be somewhat discouraging to non-formatter/binding code, but not
imply that this code is intended to be internal.


jskeet:
(Note: there's a lot of code here, but it's almost entirely just refactoring.)

docs/README.md
```diff
@@ -4,3 +4,15 @@ This directory contains documentation on:
 
 - Using the SDK as a consumer
 - Implementing new event formats and protocol bindings
+
+## Implementation utility classes
+
+The `CloudNative.CloudEvents.Impl` namespace contains utility
```
JoshLove-msft, line 10: What about Core instead of Impl?

JoshLove-msft (approved):

docs/README.md
jskeet: `Core` sounds like something that *would* be worth importing for a normal user. `Impl` isn't right either though... let's both keep thinking about it, and see if we can come up with something better. Glad you're okay with the general approach though :)

jskeet: Moved into Core now as discussed in meeting. Will merge on green.

PR #115, opened 2021-03-29
Fix folder for "Core" namespace source

jskeet:
(Previously I'd just renamed the namespace from Impl to Core, but
not moved the files.)


jskeet:
(Self-merging as this is purely a file location change.)

PR #118, opened 2021-04-14
First part of HTTP batch support

jskeet:
This proposed a general pattern (in the third commit) so it felt like it was worth getting that pattern reviewed first, before implementing it for other bindings.


jskeet:
This is odd - the tests were passing locally for me earlier, but are now failing as per CI. Will investigate.

JoshLove-msft (approved):

PR #117, opened 2021-03-31
Implement HTTP header handling properly

jskeet:
Fixes #105

This implements the spec clarification in
https://github.com/cloudevents/spec/pull/793, so we mustn't merge
this until that's been approved and merged.

Signed-off-by: Jon Skeet <jonskeet@google.com>

PR #119, opened 2021-04-15
Inline UTF-8 encoding for HTTP header values

jskeet:
This is a separate PR from #117 just so it's easy to ignore it if we want.

The first commit is from #117; the proposed change is the second commit.

If you're happy with this, I'll probably squash the second commit into #117 before merging. (I don't think we need to keep them as separate commits.)


JoshLove-msft (approved):

jskeet:
Closing, added the code to #117.

PR #120, opened 2021-04-19
Rename AspNetCore project as Http

riccardone:
to stay consistent with the other transports


jskeet:
No, I don't think this is appropriate - because it's not just general HTTP, it's specific to ASP.NET Core, in that it has ASP.NET Core dependencies. I think it's useful to make that clear in the package name.

PR #121, opened 2021-04-20
Complete the HTTP batch support

### LINKED ISSUE #44: Feature suggestion - Batched CloudEvents
**What feature would you like?**
The [CloudEvents spec](https://github.com/cloudevents/spec/blob/master/json-format.md#4-json-batch-format) describes how events could also be batched together and given a specific `Content-Type: application/cloudevents-batch+json`.

The current SDK implementation only allows single events to be parsed.

_Receiving_
It would be nice when the `DecodeJObject` takes into account if the token is an `Object` or an `Array` so we can benifit from this approach.

_Sending_
A new `HttpContent` implementation for batched events could be added for this too. Maybe take in a sequence of cloud events for this.

jskeet:
Fixes #44

Note that the other protocol bindings we have at the moment don't include batch mode. The first change prepares them to though, by rejecting application/cloudevents-batch as a content type to parse as if it represented a single CloudEvent.

### DIFF
--- a/src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs
+++ b/src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs
@@ -88,7 +88,6 @@ namespace CloudNative.CloudEvents
         }
 
         private static bool HasCloudEventsContentType(HttpRequest request) =>
-            request?.ContentType is var contentType &&
-            contentType.StartsWith(CloudEvent.MediaType, StringComparison.InvariantCultureIgnoreCase);
+            MimeUtilities.IsCloudEventsContentType(request?.ContentType);
     }
 }
--- a/src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs
+++ b/src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs
@@ -47,10 +47,12 @@ namespace CloudNative.CloudEvents
             CloudEventFormatter formatter,
             IEnumerable<CloudEventAttribute> extensionAttributes)
         {
+            Validation.CheckNotNull(httpRequest, nameof(httpRequest));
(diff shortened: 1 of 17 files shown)


src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs
```diff
@@ -87,8 +89,51 @@ public static class HttpRequestExtensions
...
+
+        /// <summary>
+        /// Converts this HTTP request into a batch of CloudEvents.
+        /// </summary>
+        /// <param name="httpRequest">The HTTP request to decode. Must not be null.</param>
+        /// <param name="formatter">The event formatter to use to process the request body. Must not be null.</param>
+        /// <param name="extensions">The extension attributes to use when populating the CloudEvent. May be null.</param>
+        /// <returns>The decoded CloudEvent.</returns>
+        /// <exception cref="ArgumentException">The request does not contain a CloudEvent.</exception>
+        public static async ValueTask<IReadOnlyList<CloudEvent>> ToCloudEventBatchAsync(
```
JoshLove-msft, line 114: Why use ValueTask here? I know sometimes the method will not be completed asynchronously, but that is only if an exception is thrown. In other worse, you won't ever get back a list of CloudEvents synchronously (or at least you will be getting back a Task from formatter.DecodeBatchModeMessageAsync)

JoshLove-msft (approved):

src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs
jskeet: Good question. Mostly because I was following the pattern for ToCloudEventAsync. I wonder whether we should make them both return Task instead. I'd like to be consistent between the two, at least.

I'll merge this PR as-is, but create an issue to remind us to discuss this.

PR #128, opened 2021-05-04
First pass at new documentation for 2.0.0-beta.2 onwards

### LINKED ISSUE #55: Clarify CloudEvent formatter responsibilities

### LINKED ISSUE #74: Revisit extensions design
I might be ignorant on this, but I haven't found any documentation to clarify regarding extensions. Right now, when you're writing a new event, I figured out how to attach extension properties using some of the extensions that are in this repo, but because of the way the Attach() method is formed, it requires some info up front in the constructor. I can set some private fields in the ctor and then reference them in the attach method. Everything builds and serializes just fine.

Coming the other direction, however... if I use `JsonEventFormatter` as suggested, it requires that I call out any extensions in the Decode method and then it seems to execute those in the exact same way which means I need those ctor arguments upfront but I don't have those because they're in the message. Short of deserizlizing to JToken and digging them out, I'm not sure how this is supposed to work.

Is this a design flaw or am I doing this wrong?

### LINKED ISSUE #98: README.md need updating for all the recent changes
Just before we release the next beta, we should update README.md, as it's currently wrong in a bunch of ways. It doesn't make sense to update it much *before* the next release though, as otherwise it's telling people how to do things with code they can't (very easily) use.

jskeet:
This is definitely a bare minimum, and we'd want more samples etc -
but it's probably enough to release 2.0.0-beta.2 and build up from
there.

Fixes #98.

Fixes #74.

Fixes #55.

### DIFF
--- a/docs/changes-since-1x.md
+++ b/docs/changes-since-1x.md
@@ -1,0 +1,79 @@
+# Changes since version 1.x
+
+Many aspects of the SDK have changed since the 1.x versions. Users
+adopting 2.x should expect to rewrite some code and retest
+thoroughly when migrating from 1.x.
+
+The following sections are not exhaustive, but describe the most
+important changes.
+
+## Core package
+
+The `CloudEvent` type constructor now only accepts the spec version
+and initial extension attributes (with no values). Everything else
+(type, ID, timestamp etc) must be set via properties or indexers.
+(In particular, the timestamp and ID are no longer populated
+automatically.) The spec version for an event is immutable once
+constructed; everything else can be modified after construction.
+
+The types used to specify attribute values must match the
+corresponding attribute type exactly; there is no implicit
+conversion available. For example, `cloudEvent["source"] =
+"https://cloudevents.io";` will fail because the `source` attribute
+is expected to be a URI.
+
+The following are now fully-abstracted concepts, rather than
+implicitly using more primitive types:
+
+- Spec version (`CloudEventSpecVersion`)
+- Attributes (`CloudEventAttribute`)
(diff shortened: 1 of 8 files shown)


docs/changes-since-1x.md
```diff
@@ -0,0 +1,79 @@
...
+- Spec version (`CloudEventSpecVersion`)
+- Attributes (`CloudEventAttribute`)
+- Attribute types (`CloudEventAttributeType`)
+
+The 1.x `CloudEventAttributes` class has now been removed, however -
+the attributes are contained directly in a map within `CloudEvent`.
+
+Timestamp attributes are now represented by `DateTimeOffset` instead
+of `DateTime`, as this provides a more specific "timestamp" concept
+with fewer amiguities.
```
JoshLove-msft, line 37:
```suggestion
with fewer ambiguities.
```

JoshLove-msft (approved):

docs/changes-since-1x.md
jskeet: Done, thanks.

PR #136, opened 2021-05-07
Introduce an attribute specifying the event formatter to use

jskeet:
This is very draft at the moment. After reviewing the concept, we'd want to:

- Consider the attribute name carefully
- Add more tests for the attribute method
- Add more tests for the `JsonEventFormatter<T>`
- Add documentation
- Add a System.Text.Json implementation of `JsonEventFormatter<T>`


jskeet:
(This is now rebased to just the four relevant commits.)

PR #153, opened 2021-05-26
Change CloudEventFormatter to use ReadOnlyMemory<byte>

jskeet:
This is only a draft, so we can discuss the impact and any better options.

### DIFF
--- a/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
@@ -65,7 +65,7 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
     /// type is absent or has a media type of "application/json", the data is parsed as JSON, with the result as
     /// a <see cref="JToken"/> (or null if the data is empty). When the content type has a media type beginning
     /// with "text/", the data is parsed as a string. In all other cases, the data is left as a byte array.
-    /// This behavior can be specialized by overriding <see cref="DecodeBinaryModeEventData(byte[], CloudEvent)"/>.
+    /// This behavior can be specialized by overriding <see cref="DecodeBinaryModeEventData(ReadOnlyMemory{byte}, CloudEvent)"/>.
     /// </para>
     /// </remarks>
     public class JsonEventFormatter : CloudEventFormatter
@@ -140,8 +140,8 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
         }
 
         /// <inheritdoc />
-        public override CloudEvent DecodeStructuredModeMessage(byte[] body, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
-            DecodeStructuredModeMessage(new MemoryStream(body), contentType, extensionAttributes);
+        public override CloudEvent DecodeStructuredModeMessage(ReadOnlyMemory<byte> body, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeStructuredModeMessage(BinaryDataUtilities.AsStream(body), contentType, extensionAttributes);
 
         /// <inheritdoc />
         public override async Task<IReadOnlyList<CloudEvent>> DecodeBatchModeMessageAsync(Stream body, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes)
@@ -164,8 +164,8 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
         }
 
         /// <inheritdoc />
-        public override IReadOnlyList<CloudEvent> DecodeBatchModeMessage(byte[] body, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
-            DecodeBatchModeMessage(new MemoryStream(body), contentType, extensionAttributes);
+        public override IReadOnlyList<CloudEvent> DecodeBatchModeMessage(ReadOnlyMemory<byte> body, ContentType contentType, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+            DecodeBatchModeMessage(BinaryDataUtilities.AsStream(body), contentType, extensionAttributes);
 
         private IReadOnlyList<CloudEvent> DecodeJArray(JArray jArray, IEnumerable<CloudEventAttribute> extensionAttributes, string paramName)
         {
@@ -352,7 +352,7 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
                 : (object) dataToken; // Deliberately cast to object to avoid any implicit conversions
 
         /// <inheritdoc />
-        public override byte[] EncodeStructuredModeMessage(CloudEvent cloudEvent, out ContentType contentType)
+        public override ReadOnlyMemory<byte> EncodeStructuredModeMessage(CloudEvent cloudEvent, out ContentType contentType)
         {
             // The cloudEvent parameter will be validated in WriteCloudEventForBatchOrStructuredMode
 
@@ -369,7 +369,7 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
         }
 
         /// <inheritdoc />
-        public override byte[] EncodeBatchModeMessage(IEnumerable<CloudEvent> cloudEvents, out ContentType contentType)
+        public override ReadOnlyMemory<byte> EncodeBatchModeMessage(IEnumerable<CloudEvent> cloudEvents, out ContentType contentType)
         {
             Validation.CheckNotNull(cloudEvents, nameof(cloudEvents));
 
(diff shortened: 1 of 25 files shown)


src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
```diff
@@ -41,5 +44,59 @@ public static byte[] ToByteArray(Stream stream)
...
+        public static byte[] ToArray(ReadOnlyMemory<byte> memory) =>
+            !MemoryMarshal.TryGetArray(memory, out var segment) || segment.Offset == 0 || segment.Count != memory.Length
+            ? memory.ToArray()
+            : segment.Array;
+
+        private static ArraySegment<byte> GetArraySegment(ReadOnlyMemory<byte> memory)
+        {
+            if (!MemoryMarshal.TryGetArray(memory, out var segment))
+            {
+                throw new ArgumentException("Unable to obtain array from memory");
```
JoshLove-msft, line 97: Interesting.. why do we throw here as opposed to falling back to ToArray?

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
```diff
@@ -41,5 +44,59 @@ public static byte[] ToByteArray(Stream stream)
...
+            await destination.WriteAsync(segment.Array, segment.Offset, segment.Count).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// TODO: Basically this is a sometimes-more-efficient version of ToArray
+        /// </summary>
+        /// <param name="memory"></param>
+        /// <returns></returns>
+        public static byte[] ToArray(ReadOnlyMemory<byte> memory) =>
+            !MemoryMarshal.TryGetArray(memory, out var segment) || segment.Offset == 0 || segment.Count != memory.Length
```
JoshLove-msft, line 89: Shouldn't this be `segment.Offset != 0`

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
```diff
@@ -41,5 +44,59 @@ public static byte[] ToByteArray(Stream stream)
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="memory"></param>
+        /// <param name="encoding"></param>
+        public static string GetString(ReadOnlyMemory<byte> memory, Encoding encoding)
+        {
+            // TODO: Use encoding.GetString(memory.Span) where we can?
```
JoshLove-msft, line 66: Good idea - looks like it is available in NetStandard 2.1.

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
JoshLove-msft: I'm not sure that it would be worth including this method.

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
jskeet: That means multi-targeting though... which adds more complexity when trying to diagnose things. Let's think about it.

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
jskeet: Yes, I messed up the condition (I'd probably reverse it in general) - but I do think it's worth having a way of getting at the data without a copy, where feasible.

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
jskeet: I'd been hoping to avoid copying as far as possible, and requiring that input data *could* be obtained as an array. We should chat about this.

jskeet:
Subject to some more discussions, this is ready for review. I've amended some aspects in terms of the "using the underlying array or creating a copy" parts of BinaryDataUtilities. I was originally somewhat reticent, but I think this is likely to be a better long-term bet than `byte[]`.

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -516,11 +515,11 @@ public override void DecodeBinaryModeEventData(byte[] body, CloudEvent cloudEven
             }
             else if (contentType.MediaType.StartsWith("text/") == true)
             {
-                cloudEvent.Data = encoding.GetString(body);
+                cloudEvent.Data = BinaryDataUtilities.GetString(body, encoding);
             }
             else
             {
-                cloudEvent.Data = body;
+                cloudEvent.Data = body.ToArray();
```
JoshLove-msft, line 522: Should we just leave it as ReadOnlyMemory?

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
```diff
@@ -15,31 +18,96 @@ public static class BinaryDataUtilities
...
+        /// <returns>The content of the stream (from its original position), as a read-only memory segment.</returns>
+        public async static Task<ReadOnlyMemory<byte>> ToReadOnlyMemoryAsync(Stream stream)
         {
             Validation.CheckNotNull(stream, nameof(stream));
-            // TODO: Optimize if it's already a MemoryStream?
+            // TODO: Optimize if it's already a MemoryStream? Will only work in some cases,
+            // and is most likely to occur in tests, where the efficiency doesn't matter as much.
             var memory = new MemoryStream();
             await stream.CopyToAsync(memory).ConfigureAwait(false);
             return memory.ToArray();
```
JoshLove-msft, line 32: Can use MemoryStream.GetBuffer here - e.g. - https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/System.Memory.Data/src/BinaryData.cs#L179

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: We *could* do, but there are some issues. My understanding is that a `ReadOnlyMemory` may be "owned" and then disposed, which would be bad news for anything still using it. (Or it could end up being a buffer that's reused later and modified, for example.) Arguably returning the original byte array was a mistake in the previous implementation.

It also makes the choice of `ReadOnlyMemory` (which isn't currently as user-friendly as `byte[]`) more visible to end-users. I'm okay with using it in `CloudEventFormatter` which is mostly used by other CloudEvents code, but I think end-user code is more likely to find byte arrays easy to use.

Feel free to argue against me though :)

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
jskeet: Except `GetBuffer` can throw `UnauthorizedException` in various cases. That's what I meant about only working in some cases :)

JoshLove-msft: Hmm good point, it looks like it will throw if constructed with the parameterless ctor, but not the one accepting a length. 
So if the stream is seekable and we construct the stream using the stream length, we should be able to use GetBuffer, which makes me think there is a bug in BinaryData.

JoshLove-msft: > That's what I meant about only working in some cases :)

I thought you were referring here to whether or not we should type sniff the stream and avoid copying. I agree that this will not always work. My suggestion is around avoiding the ToArray call, which it looks like should be possible if the original stream is seekable.

JoshLove-msft (approved):

src/CloudNative.CloudEvents/Core/BinaryDataUtilities.cs
jskeet: We can use TryGetBuffer, yes. Let's revisit this later if we need to :)

JoshLove-msft: Actually I was mistaken - GetBuffer should always work here. We own the lifetime of the MemoryStream and we are using the constructorless parameter - https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream.-ctor?view=net-5.0#System_IO_MemoryStream__ctor

jskeet: Ah - I'd misinterpreted your comment. I was hoping to use TryGetBuffer on the *incoming* stream to avoid having to copy to a new stream.
Yes, I agree we can use GetBuffer here. Will do that on Monday too, in a separate PR.

JoshLove-msft: Ah that makes sense. Yes I don't recall why we didn't use TryGetBuffer on the original stream in the BinaryData implementation - it seems like a good enhancement.

JoshLove-msft: I think it may have been related to ownership of the stream - we didn't want to have to worry about the semantics of the original stream being disposed.

PR #160, opened 2021-06-04
Improve the ASP.NET Core sample

jskeet:
- Test with both ASP.NET Core 3.1 and 5.0
- Add an API method to generate an event
- Change the launch settings to launch directly instead of with IIS Express

If #153 is merged before this, we'll need to modify it very slightly
as the encoding method will return `ReadOnlyMemory<byte>` instead of
`byte[]`, but that's easy enough to do.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,17 +17,22 @@ jobs:
     - name: Check out our repo
       uses: actions/checkout@v2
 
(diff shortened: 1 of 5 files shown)


JoshLove-msft (approved):

jskeet:
Having merged #153, this will be broken now - will rebase, fix, repush and merge on Monday. Thanks!

PR #172, opened 2021-07-08
Enable nullable reference types in all production code

### LINKED ISSUE #170: Enable c# 8.0 Nullable reference
[Nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references) helps a lot for both develop this lib and use this lib.

This project have already use c# 8.0. It will be great if it enable "Nullable reference" as well.

jskeet:
Fixes #170.

As ever, the intention is to review one commit at a time.

This is mostly pretty straightforward, but please review carefully. (At least aside from the unit test project - that doesn't matter nearly as much in terms of "theoretical" safety.)

There's one thorny part, which is the Kafka message: because we may not have a partitioning key, everything is done in terms of `Message<string?, byte[]>`... but that means the `ToCloudEvent` (etc) methods don't work when you have an actual `Message<string, byte[]>`. You can cast, of course, but it's a bit ugly.

### DIFF
--- a/src/CloudNative.CloudEvents.Amqp/AmqpExtensions.cs
+++ b/src/CloudNative.CloudEvents.Amqp/AmqpExtensions.cs
@@ -44,8 +44,8 @@ namespace CloudNative.CloudEvents.Amqp
         public static CloudEvent ToCloudEvent(
             this Message message,
             CloudEventFormatter formatter,
-            params CloudEventAttribute[] extensionAttributes) =>
-            ToCloudEvent(message, formatter, (IEnumerable<CloudEventAttribute>) extensionAttributes);
+            params CloudEventAttribute[]? extensionAttributes) =>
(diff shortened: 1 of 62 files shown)


jskeet:
Currently adding this to the unit test project as well, mostly as it can find annotation bugs...

JoshLove-msft:
What are the compat implications here? If a consuming app already had nullable refs enabled, could this potentially cause a build break?

jskeet:
> What are the compat implications here? If a consuming app already had nullable refs enabled, could this potentially cause a build break?

If you only have default settings, you'll get warnings. If you have warnings as errors, you will indeed get errors. I raised the same issue on #170. I believe the community is *generally* in favour of not considering this to be a breaking change. I think it's definitely worth doing "soon or never" though - obviously the fewer people it disrupts, the better. Happy to chat about it further.

JoshLove-msft (approved):

jskeet:
(I'll try to get a 2.1 release out next week.)

PR #177, opened 2021-07-16
"Nullable" package should be added as a development dependency

iron9light:
Based on the [doc](https://github.com/manuelroemer/Nullable#quickstart) of **Nullable** package:

This package should be added as a development dependency.

### DIFF
--- a/src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
+++ b/src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
@@ -11,7 +11,7 @@
   <ItemGroup>
     <PackageReference Include="System.Memory" Version="4.5.4" />
     <!-- Source-only package with nullable reference annotations. -->
-    <PackageReference Include="Nullable" Version="1.3.0" />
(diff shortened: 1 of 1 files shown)


jskeet:
Thanks, I'll have a look at this on Monday. (Will need to check exactly what the impact is... PrivateAssets stuff always confuses me somewhat, frankly.)

jskeet:
Hmm. Will need to look at this carefully, but it's possible that this would be a breaking change. It may be a breaking change that's worth doing anyway given that a) 2.1 hasn't been out long; b) anyone broken by it can add an explicit dependency on Nullable.

jskeet:
Okay, I've tested what happens currently on a netstandard2.0 project - and even though Nullable 1.3.0 is listed as a dependency, it doesn't actually end up being used as if it were a direct dependency. So we shouldn't break anyone by merging this, but currently folks depending on Nullable with a version lower than 1.3.0 *would* be broken.

We should merge this and release 2.1.1.

@iron9light: Please add a "Signed-off-by" line to the commit and force push, to pass the DCO check. I'll then merge and release.

jskeet (approved):

PR #181, opened 2021-08-06
Change `Uri` reference to use the `UriKind` overload in the User Guide

VisualBean:


jskeet (approved):
Fine with this as it is - I wonder whether it *might* be worth leaving one as absolute, just to make it clearer that both are okay? (The examples in https://github.com/cloudevents/spec/blob/master/spec.md#source-1 include absolute ones as well.)

VisualBean:
> https://github.com/cloudevents/spec/blob/master/spec.md#source-1

Good point. Should we leave line 237 as it was originally?

VisualBean:
> Fine with this as it is - I wonder whether it _might_ be worth leaving one as absolute, just to make it clearer that both are okay? (The examples in https://github.com/cloudevents/spec/blob/master/spec.md#source-1 include absolute ones as well.)

Good point. Should we leave line 237 as it was originally?

jskeet:
> Good point. Should we leave line 237 as it was originally?

Either that or 324 - I don't mind which.

Could you rebase your PR against the current HEAD, btw? It's possible that GitHub would just do the right thing, but I'll be more confident if the PR only has a single commit :)

VisualBean:
> > Good point. Should we leave line 237 as it was originally?
> 
> Either that or 324 - I don't mind which.
> 
> Could you rebase your PR against the current HEAD, btw? It's possible that GitHub would just do the right thing, but I'll be more confident if the PR only has a single commit :)

Yeah I noticed the weirdness and rebased.

jskeet:
Hmm... I'm still seeing 5 commits at the moment. Might it be easiest at this stage if I just create a new PR myself and merge it in?

VisualBean:
> Hmm... I'm still seeing 5 commits at the moment. Might it be easiest at this stage if I just create a new PR myself and merge it in?

Oh for crying out loud. 
Yeah maybe that will be better 😂

jskeet:
As it happens, I'd forgotten that I'm keeping the docs samples as actual source too, so the new PR will be slightly cleaner :)

PR #182, opened 2021-08-06
docs: Update one sample to show a relative URI for Source

jskeet:
(We want to make it clear that Source can be relative or absolute.)


jskeet:
Will self-merge, as this is *effectively* the change from #181, which I'd approved...

PR #189, opened 2022-01-11
Update .NET SDK and remove AppVeyor

jskeet:


jskeet:
Once this PR is approved, I'll delete the webhook for AppVeyor, which should stop it making any checks. (I'd really hoped that removing the configuration would be enough...)

JoshLove-msft (approved):

PR #190, opened 2022-01-13
Protocol bindings for AWS SNS publication, AWS SQS consumption.

Svaly:
Hello dear maintainers, 
I'm Rafal and as the title of this PR says I want to propose a protocol binding implementation for AWS Simple Notification Service and AWS Simple Queue Service.

In my company we decided to use CloudEvents for communication between applications, and since we use AWS SNS and SQS services I had the opportunity to implement this functionality. For the sake of time I did it using only CotentMode.Structured. 
I liked the work, so later I went ahead and in my spare time extended the implementation so that it fully meets the requirements for protocol bindings. So here it is, awaiting review and I hope you like it. 

Notes:
1. I know up front that I have made some questionable decisions, but I am completely open to changing or even removing them.  For example, I added integration tests that run AWS services in a docker container. The execution time for these tests is about 10sec on my machine. Also the first setup of these tests can take a very long time if the required docker image is not present in the cache, I guess this always happens in github workflow. Nevertheless, I committed these tests because I think they are helpful and give confidence that the implementation works. I would like to know your opinion. 

2. Like I said, it's already working for us, so you can take your time during the review ;)

### DIFF
--- a/CloudEvents.sln
+++ b/CloudEvents.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29001.49
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.32014.148
 MinimumVisualStudioVersion = 15.0.26124.0
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CloudNative.CloudEvents", "src\CloudNative.CloudEvents\CloudNative.CloudEvents.csproj", "{C5DC9F44-7C03-4A70-80EF-7A29696455EB}"
 EndProject
@@ -33,7 +33,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CloudNative.CloudEvents.Avr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CloudNative.CloudEvents.NewtonsoftJson", "src\CloudNative.CloudEvents.NewtonsoftJson\CloudNative.CloudEvents.NewtonsoftJson.csproj", "{9DC17081-21D8-4123-8650-D97C2153DB8C}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CloudNative.CloudEvents.SystemTextJson", "src\CloudNative.CloudEvents.SystemTextJson\CloudNative.CloudEvents.SystemTextJson.csproj", "{FACB3EF2-F078-479A-A91C-719894CB66BF}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CloudNative.CloudEvents.SystemTextJson", "src\CloudNative.CloudEvents.SystemTextJson\CloudNative.CloudEvents.SystemTextJson.csproj", "{FACB3EF2-F078-479A-A91C-719894CB66BF}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CloudNative.CloudEvents.AwsSns", "src\CloudNative.CloudEvents.AwsSns\CloudNative.CloudEvents.AwsSns.csproj", "{CAE57A21-D672-4908-9CFF-BD6E092BDC32}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CloudNative.CloudEvents.AwsSqs", "src\CloudNative.CloudEvents.AwsSqs\CloudNative.CloudEvents.AwsSqs.csproj", "{D6D46207-1DFF-4CBA-A8C2-1F64981E033B}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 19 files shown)


jskeet:
Thanks for the PR - it'll take me a while to get round to reviewing it, I'm afraid.

I don't know anything about SNS and SQS - is the way that you're using them in conjunction with CloudEvents a standardised one, or just the way that your company has decided to do it? If it's the latter, I'd suggest this would probably be a better fit for a company-specific library (which can still be open source, of course). The libraries in this repo typically implement the specifications in https://github.com/cloudevents/spec

Svaly:
> Thanks for the PR - it'll take me a while to get round to reviewing it, I'm afraid.
> 
> I don't know anything about SNS and SQS - is the way that you're using them in conjunction with CloudEvents a standardised one, or just the way that your company has decided to do it? If it's the latter, I'd suggest this would probably be a better fit for a company-specific library (which can still be open source, of course). The libraries in this repo typically implement the specifications in https://github.com/cloudevents/spec

We are using the standardized CloudEvent from this repo, I just wrote an extension to package the message serialized by the chosen formatter into AWS transport format. Later I extended the initial implementation with missing parts such as binary mode to meet the specification and share it here. 

Thanks for the link, I saw the spec and only saw a small mention of an AWS, it is about S3 event adapter, but that serves a different purpose than protocol binding. If you like my implementation, I can make an effort and propose the docs for new protocol bindings. Perhaps you could ask your friends if they would like to see such a contribution, if they don't, no problem it was already fun to write the code :)

jskeet:
Sorry for the delay - I've finally had a chance to have a closer look at this.

We definitely wouldn't take this PR exactly as it is, given that it includes a breaking change (adding an abstract method in CloudEventFormatter). But more concretely, this appears to be implicitly proposing a new binding without reference to anything that specifies that binding.

I've asked about this on the working group Zoom call, and we suggest that you create a new repo somewhere on GitHub (whether that's your personal one, a company one or in some other organization) and document the SNS/SQS binding you propose. That can then be discussed with others who are interested, and when consensus is reached there can be implementations for multiple SDKs. (I'd expect the implementations to live alongside the specification, rather than within the cloudevents GitHub org though.)

Of course, if this PR was just a "hey, I've done this thing for our internal use, and wanted to offer it back to the community" then I definitely applaud the sentiment - but please don't feel you *have* to go through all of the above. Organizing a community and getting to consensus etc is significant work, and I wouldn't suggest taking it on lightly. There's definitely potential value in standardizing this, even if only in *de facto* standardization terms. (If you can persuade Amazon to define a binding, that would be ideal of course.) But you may well find that the effort isn't worth the reward - it depends on the motivation for creating this to start with.

I'll close this PR now as I don't think we'll end up taking it in this repo anyway, but that doesn't mean the conversation has to stop.

Svaly:
Hi no problem, thank you for all the information for looking at the code and for raising question within your group.

You suggest that binding that I propose probably should be located in separate repository. Is that because it is dedicated to queue that is Amazon clod specific and for example Kafka and other bindings that you support are more like cloud agnostic?  
I feel like I'm missing something in the process, I have trouble in distinguishing what is a good fit for cloudevents GitHub org and what is not. Maybe you could explain or give a link to something that explain how it looked in the past. Why you added for example Kafka binding, you added it because Kafka is popular and many people are using it, or you started the documentation and community was asking to add Kafka, or maybe it was something else that driven this? And yes I was my sentiment to just give back to the community and maybe somebody will use it :) 

Relating to me adding an abstract method in CloudEventFormatter. I hoped you will comment this on PR so I could ask about something.
In binary mode in the formatters if DataContentType is null, a default value is used
https://github.com/cloudevents/sdk-csharp/blob/c4a938c24c3ce76bd16d2e9eeced1329c567801b/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs#L553

And later on this default value is not added as a header
https://github.com/cloudevents/sdk-csharp/blob/c4a938c24c3ce76bd16d2e9eeced1329c567801b/src/CloudNative.CloudEvents.Kafka/KafkaExtensions.cs#L158

Resulting in that during deserialization again a default value needs to be assumed 
https://github.com/cloudevents/sdk-csharp/blob/c4a938c24c3ce76bd16d2e9eeced1329c567801b/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs#L540

And the question is, wouldn't it be cleaner to always send the content type and encoding as a content-type header event if a default value was assumed, and use it for deserialization instead of assuming the default value again? The code felt more complicated because of that, for some time I could not see whole picture because the defaults was hidden from me ;d 


> Sorry for the delay - I've finally had a chance to have a closer look at this.
> 
> We definitely wouldn't take this PR exactly as it is, given that it includes a breaking change (adding an abstract method in CloudEventFormatter). But more concretely, this appears to be implicitly proposing a new binding without reference to anything that specifies that binding.
> 
> I've asked about this on the working group Zoom call, and we suggest that you create a new repo somewhere on GitHub (whether that's your personal one, a company one or in some other organization) and document the SNS/SQS binding you propose. That can then be discussed with others who are interested, and when consensus is reached there can be implementations for multiple SDKs. (I'd expect the implementations to live alongside the specification, rather than within the cloudevents GitHub org though.)
> 
> Of course, if this PR was just a "hey, I've done this thing for our internal use, and wanted to offer it back to the community" then I definitely applaud the sentiment - but please don't feel you _have_ to go through all of the above. Organizing a community and getting to consensus etc is significant work, and I wouldn't suggest taking it on lightly. There's definitely potential value in standardizing this, even if only in _de facto_ standardization terms. (If you can persuade Amazon to define a binding, that would be ideal of course.) But you may well find that the effort isn't worth the reward - it depends on the motivation for creating this to start with.
> 
> I'll close this PR now as I don't think we'll end up taking it in this repo anyway, but that doesn't mean the conversation has to stop.

jskeet:
> You suggest that binding that I propose probably should be located in separate repository. Is that because it is dedicated to queue that is Amazon clod specific and for example Kafka and other bindings that you support are more like cloud agnostic?

Yes. It's a bit of a grey area to be honest - and I haven't been involved in the CE working group for long. But the idea would be that instead of being a CloudEvents-endorsed spec directly, we'd add a link to it on the [proprietary specs page](https://github.com/cloudevents/spec/blob/v1.0.1/proprietary-specs.md). Now both of the ones links there are currently transport binding specs which are "owned" by the projects responsible for the underlying transports (Apache for RocketMQ, and Google for Google Cloud Pub/Sub). Your case would be slightly different for SNS/SQS, unless you could persuade Amazon to host the spec in their GitHub repo (which would be fabulous).

If you want the CloudEvents working group to maintain the spec in the long term, then you could *start* it in your own repo and then we could *discuss* taking over maintenance - but given that it's for an Amazon-only transport, I think it's unlikely that we'd do so.

> I feel like I'm missing something in the process, I have trouble in distinguishing what is a good fit for cloudevents GitHub org and what is not. Maybe you could explain or give a link to something that explain how it looked in the past. Why you added for example Kafka binding, you added it because Kafka is popular and many people are using it, or you started the documentation and community was asking to add Kafka, or maybe it was something else that driven this? And yes I was my sentiment to just give back to the community and maybe somebody will use it :)

I wasn't around at the time, to be honest - but I expect that Kafka being an open, widely-adopted and cross-platform system was a large part of that.

> Relating to me adding an abstract method in CloudEventFormatter. I hoped you will comment this on PR so I could ask about something.

There's a difference between structured mode and binary mode here. The CloudEvent specs don't make as clear a separate as I'd like with respect to event formatters and structured/binary mode. The [JSON format spec](https://github.com/cloudevents/spec/blob/v1.0.1/json-format.md) really only describes structured mode event handling. The [C# SDK Event Formatter docs](https://github.com/cloudevents/sdk-csharp/blob/master/docs/formatters.md) go into this in a bit more detail - but basically I didn't want to apply as much defaulting in the binary formatting as in the structured formatting.

The binary mode formatting is basically less-well-defined in the spec than structured mode formatting.

Even if that's the wrong choice, and we want to make a change, adding an abstract method like this was never going to be merged - we've shipped the SDK version 2.0, and we're not looking to make any breaking changes like that for quite some time. (We may *eventually* want a 3.0 of course, but I'm not expecting us to do that for at least a couple of years... if ever.)

Svaly:
Hi @jskeet sorry for the late reply, I love your detailed comments! Things are clearer to me now :)

I think I will change the implementation to just be a wrapper over your library and move it as you suggested to a separate repository. Maybe someone will find this interesting or helpful :)

jskeet:
So glad they're clear - I never know when I'm writing them :)

Best of luck with the project.

PR #191, opened 2022-01-18
fix: Minor tweaks to the Json.NET JsonEventFormatter for compliance

jskeet:
- Make "JSON media type detection" configurable, and default to */json and */*+json
- Default to JSON content type (including making it present in the serialized event)
- Make a "binary vs non-binary" distinction as the first part of serialization (instead of as a fallback)
- Deserialize string data values regardless of content type
- Prohibit non-string data values for non-JSON content types

### DIFF
--- a/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
@@ -30,7 +30,11 @@ namespace CloudNative.CloudEvents.NewtonsoftJson
     /// nor "data_base64" property is populated in a structured mode message.
     /// </description></item>
     /// <item><description>
-    /// If the data content type is absent or has a media type of "application/json", the data is encoded as JSON.
+    /// If the data value is a byte array, it is serialized either directly as binary data
+    /// (for binary mode messages) or as base64 data (for structured mode messages).
+    /// </description></item>
+    /// <item><description>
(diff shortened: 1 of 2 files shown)


jskeet:
(Once this is done, I'll create an equivalent PR for System.Text.Json.)

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
```diff
@@ -345,12 +359,24 @@ protected virtual void DecodeStructuredModeDataBase64Property(JToken dataBase64T
...
+        {
+            if (IsJsonMediaType(cloudEvent.DataContentType!))
+            {
+                cloudEvent.Data = dataToken;
+            }
+            else
+            {
+                if (dataToken.Type != JTokenType.String)
+                {
+                    throw new ArgumentException($"CloudEvents with a non-JSON datacontenttype can only have string data values.");
```
JoshLove-msft, line 375:
```suggestion
                    throw new ArgumentException("CloudEvents with a non-JSON datacontenttype can only have string data values.");
```

JoshLove-msft (approved):

src/CloudNative.CloudEvents.NewtonsoftJson/JsonEventFormatter.cs
jskeet: Done, thanks. Will push and merge on green.

PR #195, opened 2022-01-28
Add a Protobuf event formatter

jskeet:
(In addition, there's a first commit which fixes a validation bug discovered during implementation.)

This is complete, but there are some aspects to consider:

- Are ConvertFromProto and ConvertToProto the best names for the methods to convert between the proto and SDK representations?
- Should we add an extension method to the SDK CloudEvent and a regular method in the protobuf CloudEvent to delegate to those methods?
- Should we have a `ProtobufEventFormatter<T>` that can do packing/unpacking automatically (and probably handle the binary data mode too)?

Of these, the first should be considered before this PR is merged (although we can change it up until we do a release of course). The other two can come post-release.


src/CloudNative.CloudEvents.Protobuf/CloudNative.CloudEvents.Protobuf.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+	<TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+	<Description>Support for the Protobuf event format in for CloudNative.CloudEvents</Description>
+	<PackageTags>cncf;cloudnative;cloudevents;events;protobuf</PackageTags>
+	<LangVersion>8.0</LangVersion>
```
JoshLove-msft, line 7: can we use 9?

src/CloudNative.CloudEvents.Protobuf/ProtobufEventFormatter.cs
```diff
@@ -0,0 +1,412 @@
...
+        public override CloudEvent DecodeStructuredModeMessage(Stream messageBody, ContentType? contentType, IEnumerable<CloudEventAttribute>? extensionAttributes)
+        {
+            Validation.CheckNotNull(messageBody, nameof(messageBody));
+            return ConvertFromProto(V1.CloudEvent.Parser.ParseFrom(messageBody), extensionAttributes, nameof(messageBody));
+        }
+
+        /// <inheritdoc />
+        public override Task<CloudEvent> DecodeStructuredModeMessageAsync(Stream body, ContentType? contentType, IEnumerable<CloudEventAttribute>? extensionAttributes)
+        {
+            return base.DecodeStructuredModeMessageAsync(body, contentType, extensionAttributes);
```
JoshLove-msft, line 212: nit: await

src/CloudNative.CloudEvents.Protobuf/ProtobufEventFormatter.cs
```diff
@@ -0,0 +1,412 @@
...
+        /// its <see cref="CloudEvents.CloudEvent.Data"/> property.</param>
+        /// <param name="proto">The protobuf representation of the CloudEvent, which will be non-null.</param>
+        protected virtual void EncodeStructuredModeData(CloudEvent cloudEvent, V1.CloudEvent proto)
+        {
+            switch (cloudEvent.Data)
+            {
+                case IMessage message:
+                    proto.ProtoData = message is Any any ? any : Any.Pack(message, TypeUrlPrefix);
+                    break;
+                case string text:
```
JoshLove-msft, line 330: Does content type matter at all here?

JoshLove-msft (approved):

src/CloudNative.CloudEvents.Protobuf/CloudNative.CloudEvents.Protobuf.csproj
jskeet: We can (or even 10), but we've got 8.0 everywhere at the moment, so I'll stay consistent with that. We can change it in all projects later.

src/CloudNative.CloudEvents.Protobuf/ProtobufEventFormatter.cs
jskeet: Actually in this case I'm just going to remove this override and the batch one earlier... we shouldn't have them.

src/CloudNative.CloudEvents.Protobuf/ProtobufEventFormatter.cs
jskeet: Nope. The structured mode format captures that it's text data, so we're able to retrieve it as a string later on regardless of content type.

Issue #175, opened 2021-07-16
NuGet package is non-deterministic


jskeet:
(As shown in NuGet Package Explorer.)

This isn't a huge problem, but it's not ideal.

jskeet:
Hmm... this *still* isn't showing as deterministic. Humbug. Need to look more carefully...

PR #213, opened 2022-06-22
ci(nuget): make dotnet pack deterministic

skyfrk:
This should solve #175 according to [this blog post](https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/#deterministic-builds).

Alternatively one could also add 

```csproj
<PropertyGroup Condition="'$(GITHUB_ACTIONS)' == 'true'">
  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
</PropertyGroup>
```

to `src/Directory.Build.props`.

### DIFF
--- a/.github/workflows/nuget.yml
+++ b/.github/workflows/nuget.yml
@@ -35,6 +35,6 @@ jobs:
 
     - name: Push to NuGet
       run: |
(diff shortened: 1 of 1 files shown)


jskeet (approved):
Thanks very much for looking into this!

PR #220, opened 2022-07-01
fix: Put AvroEventFormatter in the right namespace

### LINKED ISSUE #219: AvroEventFormatter in incorrect namespace
AvroEventFormatter is currently found in the CloudNative.CloudEvents namespace.

https://github.com/cloudevents/sdk-csharp/blob/3ce4aa08f2763cfd68d5e1a0865ac9442e5b8021/src/CloudNative.CloudEvents.Avro/AvroEventFormatter.cs#L12-L16
To keep things consistent, as with both protobuf and json (first two other examples I looked at), the AvroEventFormatter should be found in CloudNative.CloudEvents.Avro namespace.

Reference to SystemTextJson and Protobuf EventFormatters below.
https://github.com/cloudevents/sdk-csharp/blob/3ce4aa08f2763cfd68d5e1a0865ac9442e5b8021/src/CloudNative.CloudEvents.Protobuf/ProtobufEventFormatter.cs#L18-L22
https://github.com/cloudevents/sdk-csharp/blob/3ce4aa08f2763cfd68d5e1a0865ac9442e5b8021/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs#L12-L16

jskeet:
We still need it in CloudNative.CloudEvents for backward
compatibility, but we can do that via derivation - which is
pleasantly simple given that the only constructor is parameterless.

Users should update to use the CloudNative.CloudEvents.Avro
namespace at their earliest convenience.

Fixes #219.

### DIFF
--- a/src/CloudNative.CloudEvents.Avro/ObsoleteFormatter.cs
+++ b/src/CloudNative.CloudEvents.Avro/ObsoleteFormatter.cs
@@ -1,0 +1,20 @@
+﻿// Copyright 2021 Cloud Native Foundation. 
+// Licensed under the Apache 2.0 license.
+// See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace CloudNative.CloudEvents;
+
+/// <summary>
+/// Formatter that implements the Avro Event Format.
+/// </summary>
+/// <remarks>
+/// This class is the wrong namespace, and is only present for backward compatibility reasons.
+/// Please use CloudNative.CloudEvents.Avro.AvroEventFormatter instead
+/// (which this class derives from for convenience).
+/// </remarks>
+[Obsolete("This class is the wrong namespace, and is only present for backward compatibility reasons. Please use CloudNative.CloudEvents.Avro.AvroEventFormatter.")]
+public class AvroEventFormatter : Avro.AvroEventFormatter
+{
+}
--- a/src/CloudNative.CloudEvents.Avro/AvroEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.Avro/AvroEventFormatter.cs
@@ -11,7 +11,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Net.Mime;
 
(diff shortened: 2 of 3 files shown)


src/CloudNative.CloudEvents.Avro/ObsoleteFormatter.cs
```diff
@@ -0,0 +1,20 @@
...
+
+/// <summary>
+/// Formatter that implements the Avro Event Format.
+/// </summary>
+/// <remarks>
+/// This class is the wrong namespace, and is only present for backward compatibility reasons.
+/// Please use CloudNative.CloudEvents.Avro.AvroEventFormatter instead
+/// (which this class derives from for convenience).
+/// </remarks>
+[Obsolete("This class is the wrong namespace, and is only present for backward compatibility reasons. Please use CloudNative.CloudEvents.Avro.AvroEventFormatter.")]
```
JoshLove-msft, line 17: I think this may be a breaking change depending on compiler options being used. Would it be sufficient to use EditorBrowsableState.Never instead?

JoshLove-msft (approved):

src/CloudNative.CloudEvents.Avro/ObsoleteFormatter.cs
jskeet: I've always taken the view that making something obsolete with a warning shouldn't be regarded as a breaking change - if users have explicitly opted in to making warnings errors, they're basically opting into having broken builds more often. There are any number of things which may introduce warnings but aren't normally considered as breaking changes. (Annoyingly, this isn't mentioned in the [canonical breaking change documentation](https://docs.microsoft.com/en-us/dotnet/core/compatibility/) as far as I can see.)

JoshLove-msft: Fair enough!

PR #224, opened 2022-07-26
feat: Allow empty payloads in Kafka, in order to support compaction tombstones

mhelleborg:
In order to support deletes for compacted topics, this PR allows empty / null payloads to be sent when using the Kafka extension.  The previous requirement that the payload is present has been dropped, and tests added.

Ref [Log compaction](https://kafka.apache.org/documentation/#compaction)

### DIFF
--- a/test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
+++ b/test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
@@ -78,6 +78,7 @@ namespace CloudNative.CloudEvents.Kafka.UnitTests
             AssertTimestampsEqual("2018-04-05T17:31:00Z", receivedCloudEvent.Time!.Value);
             Assert.Equal(MediaTypeNames.Text.Xml, receivedCloudEvent.DataContentType);
             Assert.Equal("<much wow=\"xml\"/>", receivedCloudEvent.Data);
+
             Assert.Equal("value", (string?)receivedCloudEvent["comexampleextension1"]);
         }
 
--- a/test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
+++ b/test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
@@ -133,6 +133,7 @@ namespace CloudNative.CloudEvents.Kafka.UnitTests
         [InlineData(MediaTypeNames.Application.Json, null)]
         [InlineData(MediaTypeNames.Application.Xml, "")]
         [InlineData(MediaTypeNames.Text.Plain, "")]
+        [InlineData(null, null)]
         public void KafkaBinaryMessageTombstoneTest(string contentType, object? expectedDecodedResult)
         {
             var jsonEventFormatter = new JsonEventFormatter();
--- a/test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
+++ b/test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
@@ -129,6 +129,58 @@ namespace CloudNative.CloudEvents.Kafka.UnitTests
             Assert.Equal("value", (string?)receivedCloudEvent["comexampleextension1"]);
         }
 
+        [Theory]
+        [InlineData(MediaTypeNames.Application.Octet, new byte[0])]
+        [InlineData(MediaTypeNames.Application.Json,null)]
+        [InlineData(MediaTypeNames.Application.Xml,"")]
+        [InlineData(MediaTypeNames.Text.Plain,"")]
+        public void KafkaBinaryMessageTombstoneTest(string contentType, object? expectedDecodedResult)
+        {
+            var jsonEventFormatter = new JsonEventFormatter();
+            var cloudEvent = new CloudEvent(Partitioning.AllAttributes)
+            {
+                Type = "com.github.pull.create",
+                Source = new Uri("https://github.com/cloudevents/spec/pull/123"),
+                Id = "A234-1234-1234",
+                Time = new DateTimeOffset(2018, 4, 5, 17, 31, 0, TimeSpan.Zero),
+                DataContentType = contentType,
+                Data = null,
+                ["comexampleextension1"] = "value",
+                [Partitioning.PartitionKeyAttribute] = "hello much wow"
+            };
+
+            var message = cloudEvent.ToKafkaMessage(ContentMode.Binary, new JsonEventFormatter());
+            Assert.True(message.IsCloudEvent());
+            
+            // Sending an empty message is equivalent to a delete (tombstone) for that partition key, when using compacted topics in Kafka.
+            // This is the main use case for empty data messages with Kafka.
+            Assert.Empty(message.Value);
+
+            // Using serialization to create fully independent copy thus simulating message transport.
+            // The real transport will work in a similar way.
+            var serialized = JsonConvert.SerializeObject(message, new HeaderConverter());
+            var settings = new JsonSerializerSettings
+            {
+                Converters = { new HeadersConverter(), new HeaderConverter() }
+            };
+            var messageCopy = JsonConvert.DeserializeObject<Message<string?, byte[]>>(serialized, settings)!;
+
+            Assert.True(messageCopy.IsCloudEvent());
+            var receivedCloudEvent = messageCopy.ToCloudEvent(jsonEventFormatter, Partitioning.AllAttributes);
+
+            Assert.Equal(CloudEventsSpecVersion.Default, receivedCloudEvent.SpecVersion);
+            Assert.Equal("com.github.pull.create", receivedCloudEvent.Type);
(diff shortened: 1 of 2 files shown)


jskeet:
I'll have a look at this when I get the chance, but I'm afraid it may be a while due to vacations. (I don't know anything about Kafka, which means it'll take me longer to try to work out whether I think it's a good idea or not...)

mhelleborg:
Thanks Jon :)

The change should make it easy to use CloudEvents for projecting state using compacted topics, as a streaming KV-store. 

The zero-length payloads are treated as tombstones, and lets Kafka clean up removed messages over time.

jskeet:
I still don't know why we prohibited empty messages in the first place. I'm going to ask at the meeting tomorrow. If other SDKs are fine with empty payloads, we should be too. If they're not, I want to find out why :)

pierDipi:
Commenting after the WG discussion.

I agree with this PR.

In general, Kafka records with a `null` value are valid records with both [log compacted topics](https://kafka.apache.org/documentation/#compaction) or with regular topics.

However, in the log compacted topics case, a record with a `null` value has a different semantic (tombstone), they signal to Kafka that it can remove all previous records with the same key but *it retains and delivers* to subscribers the tombstone record as long as it is consumed before the [the topic-level configuration `delete.retention.ms`](https://kafka.apache.org/20/documentation.html#topicconfigs) expires.

Since configuring a topic to be log compacted is an event producer's responsibility, the producer is specifically asking for that behavior so I don't see why it shouldn't be allowed to send a CloudEvent that is a tombstone given that:
- the event itself is just a normal CloudEvent
- the event producer is specifically asking the transport to have that side effect

jskeet:
Just to set expectations, I'm hoping to merge this PR next week - I'm on vacation tomorrow, and done with work for today.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
```diff
@@ -129,6 +129,58 @@ public void KafkaBinaryMessageTest()
...
+        public void KafkaBinaryMessageTombstoneTest(string contentType, object? expectedDecodedResult)
+        {
+            var jsonEventFormatter = new JsonEventFormatter();
+            var cloudEvent = new CloudEvent(Partitioning.AllAttributes)
+            {
+                Type = "com.github.pull.create",
+                Source = new Uri("https://github.com/cloudevents/spec/pull/123"),
+                Id = "A234-1234-1234",
+                Time = new DateTimeOffset(2018, 4, 5, 17, 31, 0, TimeSpan.Zero),
+                DataContentType = contentType,
```
jskeet, line 146: While not strictly invalid as far as I'm aware, is there any good reason to have a data content type when there isn't any data?

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
```diff
@@ -129,6 +129,58 @@ public void KafkaBinaryMessageTest()
             Assert.Equal("value", (string?)receivedCloudEvent["comexampleextension1"]);
         }
 
+        [Theory]
+        [InlineData(MediaTypeNames.Application.Octet, new byte[0])]
+        [InlineData(MediaTypeNames.Application.Json,null)]
```
jskeet, line 134: Nit: please add a space after the comma here and below.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
```diff
@@ -129,6 +129,58 @@ public void KafkaBinaryMessageTest()
...
+
+            Assert.True(messageCopy.IsCloudEvent());
+            var receivedCloudEvent = messageCopy.ToCloudEvent(jsonEventFormatter, Partitioning.AllAttributes);
+
+            Assert.Equal(CloudEventsSpecVersion.Default, receivedCloudEvent.SpecVersion);
+            Assert.Equal("com.github.pull.create", receivedCloudEvent.Type);
+            Assert.Equal(new Uri("https://github.com/cloudevents/spec/pull/123"), receivedCloudEvent.Source);
+            Assert.Equal("A234-1234-1234", receivedCloudEvent.Id);
+            AssertTimestampsEqual("2018-04-05T17:31:00Z", receivedCloudEvent.Time!.Value);
+            Assert.Equal(contentType,receivedCloudEvent.DataContentType);
```
jskeet, line 176: Nit: please add a space after the comma.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
```diff
@@ -129,6 +129,58 @@ public void KafkaBinaryMessageTest()
...
+            Assert.Equal(expectedDecodedResult, receivedCloudEvent.Data);
+            Assert.Equal("hello much wow", (string?) receivedCloudEvent[Partitioning.PartitionKeyAttribute]);
+
+            Assert.Equal("value", (string?)receivedCloudEvent["comexampleextension1"]);
+        }
+
```
jskeet, line 182: Nit: please remove one of these blank lines.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
jskeet: (It looks like you're using that to effectively check the result of decoding different empty bodies - I don't think that's really part of the Kafka code, so I think you can probably simplify this test. If you'd prefer to keep it, that's okay.)

jskeet (approved):
Approved in general, with a few nits.

I'm going to create a PR for the Kafka binding spec at https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/kafka-protocol-binding.md to mention that an empty payload can have a special meaning when using binary mode.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
mhelleborg: I was thinking it's a good idea to verify that the deserialization is well behaved across all the content types, even without a payload.

I can add the null content type as well, to be sure that it works correctly too.

jskeet: But isn't that deserialization purely in the formatter, not in the Kafka code at all?

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
jskeet: This has been resolved, but there's still an extra blank line as far as I can see...

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
```diff
@@ -156,17 +208,16 @@ public override bool CanConvert(Type objectType)
                 {
                     return null;
                 }
-                else 
+
```
jskeet, line 211: Is there any actual change to the code here? It seems unrelated to the PR.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
```diff
@@ -78,7 +78,6 @@ public void KafkaStructuredMessageTest()
             AssertTimestampsEqual("2018-04-05T17:31:00Z", receivedCloudEvent.Time!.Value);
             Assert.Equal(MediaTypeNames.Text.Xml, receivedCloudEvent.DataContentType);
             Assert.Equal("<much wow=\"xml\"/>", receivedCloudEvent.Data);
-
```
jskeet, line 81: I'm not sure why this line has been removed, but there's a blank line in the equivalent assertion that you've added at (new) line 128.

It would be good if this PR *only* contained changes that were actually relevant to the feature.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
mhelleborg: True, it might be a case of over-testing. But I saw it as a better safe than sorry to cover it. 

I can remove the content type checks if you wish?

jskeet: I don't mind too much - it just seems unnecessary. I haven't looked at what the coverage is like at the moment.

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
mhelleborg: Purely structure. Should I revert to if / else instead of guard?

test/CloudNative.CloudEvents.UnitTests/Kafka/KafkaTest.cs
jskeet: This conversation has been resolved, but I still see the change... again, I think the PR would be best if it were *just* changes related to the feature. (Other formatting changes can come in a separate PR, of course.)

PR #240, opened 2022-10-10
Update dependencies for CloudNative.CloudEvents.AspNetCore

### LINKED ISSUE #239: Transitive dependencies have high/critical security vulnerabilities
From running `dotnet list package --vulnerable --include-transitive` in CloudNative.CloudEvents.AspNetCore:

```text
Project `CloudNative.CloudEvents.AspNetCore` has the following vulnerable packages
   [netstandard2.0]:
   Transitive Package               Resolved   Severity   Advisory URL
   > Newtonsoft.Json                9.0.1      High       https://github.com/advisories/GHSA-5crp-9r3c-p9vr
   > System.Text.Encodings.Web      4.5.0      Critical   https://github.com/advisories/GHSA-ghhp-997w-qr28

   [netstandard2.1]:
   Transitive Package               Resolved   Severity   Advisory URL
   > Newtonsoft.Json                9.0.1      High       https://github.com/advisories/GHSA-5crp-9r3c-p9vr
   > System.Text.Encodings.Web      4.5.0      Critical   https://github.com/advisories/GHSA-ghhp-997w-qr28
```

Both of these come from Microsoft.Extensions.DependencyModel 2.1.0. There's no upgrade in the 2.x line for this. We could potentially upgrade just Newtonsoft.Json and System.Text.Encodings.Web, but we don't really want a Newtonsoft.Json dependency at all.

Upgrading to 3.1.25 would fix this. It does contain a breaking change: `DependencyContextJsonReader.ReadTargetLibraryDependencies` has been removed.

It looks like we don't actually use the Mvc.Core dependency anyway. Dropping that would fix the Newtonsoft.Json issue, and we can update System.Text.Encodings.Web explicitly.

jskeet:
Fixes #239

### DIFF
--- a/src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
+++ b/src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
@@ -9,8 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.22" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.1.16" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.34" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0" />
     <ProjectReference Include="..\CloudNative.CloudEvents\CloudNative.CloudEvents.csproj" />
   </ItemGroup>
 


jskeet:
Alternative: we could update to System.Text.Encodings.Web version 4.5.1 or 4.7.1, if that would be better than going straight to 6.0.0.

mattjohnsonpint (approved):

mattjohnsonpint:
This will fix the issue.  Thanks.

Regarding:

> Alternative: we could update to System.Text.Encodings.Web version 4.5.1 or 4.7.1, if that would be better than going straight to 6.0.0.

While these version numbers now appear to be following the .NET release cycle, it appears 6.0.0 is targeting `net6.0;netcoreapp3.1;netstandard2.0;net461`
https://www.nuget.org/packages/System.Text.Encodings.Web/6.0.0#supportedframeworks-body-tab

That should be compatible with this library, which is targeting `netstandard2.0;netstandard2.1`.


Separately, you might want to consider whether it's worth expanding the targets.  The latest guidance is [here](https://learn.microsoft.com/dotnet/standard/library-guidance/cross-platform-targeting), However, since .NET Framework 4.6.1 and .NET 5.0 are out of support, perhaps it would be more appropriate to do something like: `net6.0;netstandard2.0;netstandard2.1;net462` in this library and upstream?

jskeet:
Thanks for the review. Will get @JoshLove-msft to have a look too, then merge and release. (I'll do a full minor version release across all packages just for the sake of simple consistency when it's merged.)

> Separately, you might want to consider whether it's worth expanding the targets.

I'd rather not - that way we'd have to keep updating the platforms as the set of supported frameworks changes. If we only need APIs within netstandard, I'd prefer to stick to just those. We can add additional platforms if/when we really need them.

JoshLove-msft (approved):

jskeet:
Thanks Josh. Will do a release on Monday.

JoshLove-msft:
> Thanks Josh. Will do a release on Monday.

Sounds good!

Issue #256, opened 2023-03-03
Unexpected auto-generated class name from CloudNative.CloudEvents.Protobuf 


gbrulotte:
Hi, we have a use case where we want to generate a CloudEvent in a service using CloudNative.CloudEvents and send it in the protobuf format using CloudNative.CloudEvents.Protobuf in a gRPC call. We defined a custom .proto containing the service and wanted to re-use the existing CloudEvent proto (https://github.com/cloudevents/spec/blob/main/cloudevents/formats/cloudevents.proto) within our service proto to be able to deserialize the event on the server side in a CloudNative.CloudEvents CloudEvent.

Since the CloudNative nuget package already bundles the auto-generated classes, we need to reference the CloudEvent proto without compiling it. With this configuration, I get this error:

`error CS0234: The type or namespace name 'CloudeventsReflection' does not exist in the namespace 'CloudNative.CloudEvents.V1' (are you missing an assembly reference?)
`

From what I see in https://github.com/cloudevents/sdk-csharp/blob/main/generate_protos.sh, the cloudevents.proto is copied in a file named ProtoSchema.proto. This results in an auto-generated class named ProtoSchemaReflection (https://github.com/cloudevents/sdk-csharp/blob/main/src/CloudNative.CloudEvents.Protobuf/ProtoSchema.g.cs#L15) instead of being named CloudeventsReflection.

Would it be possible to instead keep the original proto file name (cloudevents.proto) to keep compatibility?

Here is a sample showing the problem: https://github.com/gbrulotte/CloudEventProtobufSample.

Thanks!

jskeet:
Unfortunately changing that now would be a breaking change for anyone who already refers to ProtoSchemaReflection. I suspect there was some problem with keeping the original name - I don't know why else I'd have changed it, to be honest.

I'll have a closer look on Monday.

jskeet:
Will try to look fully later on today, but one option I'll look into is renaming, but adding a handwritten ProtoSchemaReflection class which just delegates to CloudEventReflection, for compatibility.

jskeet:
Will do a release tomorrow.

jskeet:
Release 2.6.0 is now available - do let me know if you still have issues. (I've tried your sample project, and that now builds, so hopefully that was all that was needed.)

Issue #252, opened 2023-01-16
Embed license on the nuget package


mieliespoor:
The [nuspec specification](https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg) allows for the license to be specified within the nuspec/csproj files to make it easier to determine which license is being used by the package. 

Could all the package from this repo add this to help making it easier for consumers?

`<license type="expression">Apache-2.0</license>`

jskeet:
Possibly. I know we *don't* do that for Google libraries - there's some disadvantage to using the SPDY identifier instead of embedding the license file directly. I'll look into it; it may not be relevant to CNCF libraries.

jskeet:
Have asked about this. A conservative reading of the license suggests that the license really should be embedded in the package. Unfortunately we can't specify both `<PackageLicenseFile>` and `<PackageLicenseExpression>` in the csproj file (or Directory.Build.props in this particular case).

What I might do is embed the licence file as a regular file within the package, so that we *are* shipping it, but use `<PackageLicenseExpression>` to make the package licensing more machine-readable. Will try to get to that this week.

PR #258, opened 2023-03-07
Release 2.6.0

jskeet:
Changes since 2.5.1:

- Dependencies: system-level dependencies updated
- The NuGet package now uses PackageLicenseExpression (but still includes the licence file as well). Fixes ([#252](https://github.com/cloudevents/sdk-csharp/issues/252)).
- Regenerated protobuf schema using the original filename of cloudevents.proto instead of ProtoSchema.proto. An additional ProtoSchemaReflection class has been added purely for compatibility. Fixes ([#256](https://github.com/cloudevents/sdk-csharp/issues/256)).

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -1,5 +1,16 @@
 # Version history (from 2.0)
 
+## 2.6.0 (2023-03-07)
+
+- Dependencies: system-level dependencies updated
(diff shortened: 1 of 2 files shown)


jskeet:
(Will merge on green and trigger the release.)

PR #265, opened 2023-08-31
Prevent string resource lookup if not required

### LINKED ISSUE #264: String resource is looked up eagerly for validation methods
This causes an unnecessary string resource lookup each time the indexer is called

https://github.com/cloudevents/sdk-csharp/blob/0c0169f4fd94439b1dfbf4e14f2b004f785b0b21/src/CloudNative.CloudEvents/CloudEvent.cs#L125

https://github.com/cloudevents/sdk-csharp/blob/0c0169f4fd94439b1dfbf4e14f2b004f785b0b21/src/CloudNative.CloudEvents/CloudEvent.cs#L139

https://github.com/cloudevents/sdk-csharp/blob/0c0169f4fd94439b1dfbf4e14f2b004f785b0b21/src/CloudNative.CloudEvents/CloudEvent.cs#L184

https://github.com/cloudevents/sdk-csharp/blob/0c0169f4fd94439b1dfbf4e14f2b004f785b0b21/src/CloudNative.CloudEvents/CloudEvent.cs#L190

campbellhardingdeason:
Fixes #264

### DIFF
--- a/src/CloudNative.CloudEvents/Core/Validation.cs
+++ b/src/CloudNative.CloudEvents/Core/Validation.cs
@@ -46,7 +46,7 @@ namespace CloudNative.CloudEvents.Core
         /// <param name="condition">The condition to validate; this method will throw an <see cref="ArgumentException"/> if this is false.</param>
         /// <param name="paramName">The name of the parameter being validated. May be null.</param>
         /// <param name="messageFunc">A func that returns the message to use in the exception, if one is thrown.</param>
-        public static void CheckArgument([DoesNotReturnIf(false)] bool condition, string paramName, Func<string> messageFunc)
+        internal static void CheckArgument([DoesNotReturnIf(false)] bool condition, string paramName, Func<string> messageFunc)
         {
             if (!condition)
             {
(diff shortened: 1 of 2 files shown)


src/CloudNative.CloudEvents/Core/Validation.cs
```diff
@@ -40,6 +40,20 @@ public static void CheckArgument([DoesNotReturnIf(false)] bool condition, string
             }
         }
 
+        /// <summary>
+        /// Validates an argument-dependent condition, throwing an exception if the check fails.
+        /// </summary>
+        /// <param name="condition">The condition to validate; this method will throw an <see cref="ArgumentException"/> if this is false.</param>
+        /// <param name="paramName">The name of the parameter being validated. May be null.</param>
+        /// <param name="messageFunc">A func that returns the message to use in the exception, if one is thrown.</param>
+        public static void CheckArgument([DoesNotReturnIf(false)] bool condition, string paramName, Func<string> messageFunc)
```
jskeet, line 49: Could we keep this internal to start with? It looks like we only *need* it to be internal for now - we're not using resource strings elsewhere, and I suspect other users won't be doing so either, so I'd *prefer* to avoid exposing it.

jskeet (commented):
Thanks very much for spotting this, and for the PR. Just one small requested change, if you're happy to make it.

src/CloudNative.CloudEvents/Core/Validation.cs
campbellhardingdeason: Sure thing 👍

jskeet (approved):
Thanks - I'll squash and merge down into one commit.
I'll try to get a release done next week.

Issue #278, opened 2024-02-04
Only include System.Memory for Net Standard 2.0


thompson-tomo:
### Summary
I want System.Memory to only be a dependency when not provided by the application framework

### Proposal
Make System.Memory added conditionally so that it is only included for Net Standard 2.0

jskeet:
Yup, that sounds reasonable (although I have vague memories of trying it when we added another framework, and failing; I'm hoping I'm just misremembering.)

thompson-tomo:
@jskeet i am more than happy to take a look at given i have done this on other projects

jskeet:
PRs definitely welcome :) I'm on vacation at the moment though, so may take a little while to review.

thompson-tomo:
No problem at all & I am in no rush. Pr is done as  #279 and note I found 2 other packages I could tweak the conditions on to reduce dependencies, so I did them at the same time.

PR #279, opened 2024-02-04
#278 Only include packages not part of framework

thompson-tomo:
### Summary
Add additional logic to dependencies so that only necessary dependencies are added

### Details
Add additional conditions so that the dependency graph comes down for the newer frameworks

Closes

- #278

### DIFF
--- a/src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
+++ b/src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='netstandard2.1'" />
+    <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)'=='netstandard2.0'" />
     <!-- Source-only package with nullable reference annotations. -->
     <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="All" />
   </ItemGroup>
--- a/src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
+++ b/src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.34" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0" Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='netstandard2.1'" />
     <ProjectReference Include="..\CloudNative.CloudEvents\CloudNative.CloudEvents.csproj" />
   </ItemGroup>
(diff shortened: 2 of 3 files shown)


src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='netstandard2.1'" />
+    <PackageReference Include="System.Memory" Version="4.5.5" Condition="'$(TargetFramework)'=='netstandard2.0'" />
```
jskeet, line 12: Might this one still be useful? The current version was released in May 2022, which is after .NET 6 was released... so it may contain bugfixes that are useful?

jskeet (commented):
Generally feels good to me, but I'm going to ask for a second opinion.

jskeet:
@captainsafia please could you have a look at this? I suspect it's the kind of thing you live and breathe...

src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
thompson-tomo: No issue in re-adding it if you would like. Alternative suggestion, do we even need net-standard 2.1 (note I have not checked for framework conditional code).

thompson-tomo: Fine for me. Don't see a need to spend too much time on the framework let me know the preference ie re-add for net standard 2.1 or leave it as proposed.

jskeet: Let's re-add netstandard2.1 - @captainsafia can always disagree :)

thompson-tomo: Ok will re-add it tomorrow morning my time (ie in 10hrs), if I haven't been told otherwise.

captainsafia: From a strictly compatibility perspective, we're probably good to just target `netstandard2.0`. Although it appears that the `netstandard2.1` target was originally added to support taking advantage of the nullability annotations that come built-in with `netstandard2.1`, I think the fact that we also target `net8.0` now means we have good coverage here.

A quick peruse of the code doesn't reveal that we're using any APIs that are only available in `netstandard2.1` and not `netstandard2.0` so I think we can remove it.

jskeet: Thanks @captainsafia. I'll try installing the package as a dependency of both a netstandard2.0 and netstandard2.1 project tomorrow, to see if there are any usability differences.

(I'd be happy for us to remove it in a separate PR if that makes anything easier, btw. We could get this one merged first, if that's useful.)

captainsafia (approved):
Conditional references look good!

Agree with @jskeet about leaving the question of whether we target `netstandard2.1` at all now that we have `net8.0` targets to a follow-up PR.

src/CloudNative.CloudEvents/CloudNative.CloudEvents.csproj
jskeet: Great. As we may remove netstandard2.1 anyway, I'll approve this and merge without you adding the condition back in, @thompson-tomo :)

jskeet (approved):

PR #287, opened 2024-03-20
Optimize encoding

bencyoung-Fignum:
This change optimizes the binary mode formatter when UTF8 is chosen, and allows the memory stream implementation to be overridden (e.g. RecyclableMemoryStream)


jskeet:
Will try to look at this soon, but it may be next week.

jskeet:
There are a few things I'd probably want to change here - I wonder if the best thing is for me to create an PR from scratch rather than leave a bunch of nitpicks, then get you to review that. Does that work for you?

bencyoung-Fignum:
Absolutely, it was just a PoC

jskeet:
Great. I'll try to come up with something next week.

PR #289, opened 2024-03-22
#286 Add in a framework reference for net8

### LINKED ISSUE #286: Move to a framework reference
The CloudNative.CloudEvents.AspNetCore package currently has a dependency on the Microsoft.AspNetCore.Http nuget package which was deprecated as part of the dotnet clean up https://github.com/dotnet/announcements/issues/217.

As such we should move to using a framework reference to eliminate this dependency especially as the name clearly indicates it is for aspnet.

thompson-tomo:
This adds in the asp net framework reference for the net8 TFM so that packages can be decreased.

Closes #286


jskeet:
The build is failing, presumably due to some differences in nullability annotations. Are you happy to look into that yourself, or would you like me to do it in a separate commit still within this PR?

thompson-tomo:
Yeap, I will look through it over the next couple of days.

thompson-tomo:
Have gone through and fixed the null reference errors and at the same time updated the tests based https://github.com/aspnet/Announcements/issues/377 which was being used in a number of locations.

jskeet:
Will try to look at this next week. (Today and Monday are UK public holidays.)

thompson-tomo:
No problem at all. 🙂

src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
```diff
@@ -8,9 +8,13 @@
     <PackageTags>cncf;cloudnative;cloudevents;events;aspnetcore;aspnet</PackageTags>
   </PropertyGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.34" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="6.0.0" Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='netstandard2.1'" />
+    <ItemGroup>
```
jskeet, line 11: Could all of these just be in a single ItemGroup? Alternatively, maybe create one ItemGroup for netstandard2.0/2.1, and one for everything else, so we don't need a condition on each element within it.

test/CloudNative.CloudEvents.UnitTests/AspNetCore/HttpResponseExtensionsTest.cs
```diff
@@ -114,11 +114,17 @@ public async Task CopyToHttpResponseAsync_Batch()
...
             AssertBatchesEqual(batch, parsedBatch);
         }
 
-        private static HttpResponse CreateResponse() => new DefaultHttpResponse(new DefaultHttpContext()) { Body = new MemoryStream() };
+        private static HttpResponse CreateResponse()
+        {
+            var response = new DefaultHttpContext().Response;
+            response.Body = new MemoryStream();
+            return response;
+        }
```
jskeet, line 127: Nit: please add a lack line after this method.

jskeet (commented):
A couple of nits, but I'm happy otherwise. @captainsafia please could you review as well?

test/CloudNative.CloudEvents.UnitTests/AspNetCore/HttpResponseExtensionsTest.cs
thompson-tomo: Will do

src/CloudNative.CloudEvents.AspNetCore/CloudNative.CloudEvents.AspNetCore.csproj
thompson-tomo: I usually do item group as well but was following the existing packages. Will update tomorrow

thompson-tomo:
Agree with the 2 proposed changes & I will make them other the weekend.

captainsafia (approved):

jskeet (approved):
Great, thanks! Will squash and merge.

PR #305, opened 2024-08-13
Remove SourceLink package reference

### LINKED ISSUE #304: Remove SourceLink?
Now that we're using the .NET 8 SDK, we probably don't need it at all.

jskeet:
With .NET 8, SourceLink is available by default.

Fixes #304.

cc @captainsafia

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -8,9 +8,8 @@
     <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)"/>
   </ItemGroup>
   
-  <!-- Use SourceLink for all production projects, and include the licence file -->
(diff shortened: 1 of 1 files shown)


captainsafia (approved):
Looks good!

While we're thinking about features we're taking advantage of in the .NET 8 SDK, should we update the `globa.json` to point to `8.0.400`?

In case you're interested in automating this, there's [a GitHub action](https://github.com/marketplace/actions/update-net-sdk) we use on the ASP.NET Core repo to manage automatic updates.

jskeet:
Thanks - I'll create a new PR (and merge it) to update to 8.0.400, then look at setting up Renovate - that definitely handles global.json as well as other dependencies. (I assume it knows about Directory.Packages.props...)

Issue #310, opened 2024-09-17
CloudNative.CloudEvents.RabbitMq


antiblue:
Hi, I worked on an AMQP 0.9.1 protocol binding using the [RabbitMq.Client](https://www.rabbitmq.com/client-libraries/dotnet-api-guide#overview).

I followed the ["RabbitMQ Protocol Binding for CloudEvents" specification](https://github.com/knative-extensions/eventing-rabbitmq/blob/main/cloudevents-protocol-spec/spec.md)
and looked (of course) at how the [AMQP 1.0 bindings](https://github.com/cloudevents/sdk-csharp/tree/main/src/CloudNative.CloudEvents.Amqp) here do things. 

I would like to create a PR to get this integrated into the sdk-csharp.

jskeet:
That sounds fine, yes. Maybe leave this issue open for any discussion about aspects that there isn't code for yet, and create a PR to add the library (and tests).

jskeet:
(If the new library can be independent of the existing Amqp library, that would be handy in terms of diamond dependencies etc, but let me know if that doesn't work out. I really don't know a lot about this area I'm afraid.)

antiblue:
Yes, I could not use the existing library and the new one does not depend on `CloudNative.CloudEvents.Amqp` or `AMQPNetLite`

antiblue:
Hi @jskeet I face a strange issue.
I am unable to restore packages on [my branch](https://github.com/antiblue/sdk-csharp/tree/feature/T-310-rabbitmq) because he basically tries to find itself.

The error message is
`[NU1101] Unable to find package CloudNative.CloudEvents.RabbitMq. No packages exist with this id in source(s): Microsoft Visual Studio Offline Packages, nuget.org at (0:0)`

Any ideas?

jskeet:
I'll see if I can find time to pull your branch and reproduce, but I make no guarantees I'm afraid.

jskeet:
Got it - the problem is that it's trying to compare the current code with a previous version of the same package (for backward compatibility). Add this into the property group:

```xml
<EnablePackageValidation>false</EnablePackageValidation>
```

... then do a full rebuild. It should work at that point.

PR #311, opened 2024-09-19
Adds RabbitMQ bindings

antiblue:
PR for #310 
Adds AMQP 0.9.1 protocol binding using the [RabbitMq.Client](https://www.rabbitmq.com/client-libraries/dotnet-api-guide#overview).

It follows the ["RabbitMQ Protocol Binding for CloudEvents" specification](https://github.com/knative-extensions/eventing-rabbitmq/blob/main/cloudevents-protocol-spec/spec.md).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -333,3 +333,6 @@ ASALocalRun/
 
 # Local History for Visual Studio
 .localhistory/
+
+# Jetbrains Rider cache/options directory
+.idea/
--- a/CloudEvents.sln
+++ b/CloudEvents.sln
@@ -72,6 +72,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "xml", "xml", "{4012C753-68D
 		conformance\format\xml\valid-events.xml = conformance\format\xml\valid-events.xml
 	EndProjectSection
 EndProject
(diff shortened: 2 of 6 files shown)


jskeet:
Will try to review this next week.

clemensv:
There is no spec'd protocol binding for AMQP 0.9.1. AMQP 0.9.1 is an expired draft of the AMQP protocol and never made it to standardization. That excludes AMQP 0.9.1 per https://github.com/cloudevents/spec/blob/main/cloudevents/primer.md#qualifying-protocols-and-encodings 

What can be done is to submit a binding spec into the "proprietary specs" section for RabbitMQ analog to RocketMQ. 
https://github.com/cloudevents/spec/blob/main/cloudevents/proprietary-specs.md

That can then be composed with, but arguably not embedded into the official SDK.

antiblue:
> Edit: I only just see that the RMQ binding exists but isn't registered in our proprietary spec list. That should be added. Still skeptical whether a proprietary client should be in the offical SDK.

That is a valid point.

antiblue:
IMHO Clemens arguments are a reason to close this pull request.

jskeet:
Thanks, doing so now.

Issue #313, opened 2024-09-25
Kafka Cloud events implementation does not provide support For Null, Ignore Key types


Abuntxa:
The current extensión methods that allow transforming to and from CloudEvents into Kafka messages expects always messages of type Message<string?, byte[]> where the first generic template parameter is the type of Key in the Kafka messages.

When using the KafkaConsumer or producer it is usual to use the types "Confluent.Kafka.Ignore" or "Confluent.Kafka.Null" too that would handle messages like the following:
Message<Confluent.Kafka.Ignore, byte[]>
Message<Confluent.Kafka.Null, byte[]>

jskeet:
This is something I would be unable to take on myself, not being familiar with Kafka. (I basically inherited the current code to maintain.) I'd welcome a pull request that addresses the concern, preferably without any breaking changes to the existing API.

jskeet:
Closing, at least for now, as the PR was closed.

PR #314, opened 2024-09-26
feat: provide support for Kafka message keys different than "string?"

Abuntxa:
This pull request implements support for [Kafka Message key types other than "string?".](https://github.com/cloudevents/sdk-csharp/issues/313)

### DIFF
--- a/src/CloudNative.CloudEvents.Kafka/KafkaExtensions.cs
+++ b/src/CloudNative.CloudEvents.Kafka/KafkaExtensions.cs
@@ -1,9 +1,10 @@
-﻿// Copyright (c) Cloud Native Foundation.
+// Copyright (c) Cloud Native Foundation.
 // Licensed under the Apache 2.0 license.
 // See LICENSE file in the project root for full license information.
 
 using CloudNative.CloudEvents.Core;
 using CloudNative.CloudEvents.Extensions;
+using CloudNative.CloudEvents.Kafka.PartitionKeyAdapters;
 using Confluent.Kafka;
 using System;
 using System.Collections.Generic;
@@ -32,7 +33,7 @@ namespace CloudNative.CloudEvents.Kafka
         /// </remarks>
         /// <param name="message">The message to check for the presence of a CloudEvent. Must not be null.</param>
         /// <returns>true, if the request is a CloudEvent</returns>
-        public static bool IsCloudEvent(this Message<string?, byte[]> message) =>
+        public static bool IsCloudEvent<TKey>(this Message<TKey, byte[]> message) =>
             GetHeaderValue(message, SpecVersionKafkaHeader) is object ||
             MimeUtilities.IsCloudEventsContentType(GetHeaderValue(message, KafkaContentTypeAttributeName));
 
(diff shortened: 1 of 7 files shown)


src/CloudNative.CloudEvents.Kafka/PartitionKeyAdapters/BinaryGuidPartitionKeyAdapter.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+
+namespace CloudNative.CloudEvents.Kafka.PartitionKeyAdapters
```
jskeet, line 3: Please use file-scoped namespaces for new code. (It reduces the indentation level.)

Additionally, please add a copyright notice to each new file (just copy it from existing files).

jskeet (commented):
A couple of comments to streamline changes later, but I won't have time to review the actual code until next week.

src/CloudNative.CloudEvents.Kafka/KafkaExtensions.cs
```diff
@@ -32,7 +33,7 @@ public static class KafkaExtensions
         /// </remarks>
         /// <param name="message">The message to check for the presence of a CloudEvent. Must not be null.</param>
         /// <returns>true, if the request is a CloudEvent</returns>
-        public static bool IsCloudEvent(this Message<string?, byte[]> message) =>
+        public static bool IsCloudEvent<TKey>(this Message<TKey, byte[]> message) =>
```
jskeet, line 36: Note: this is a binary breaking change. (We'll need to add a generic overload to the existing non-generic method.)

src/CloudNative.CloudEvents.Kafka/PartitionKeyAdapters/BinaryGuidPartitionKeyAdapter.cs
Abuntxa: Corrected as sugested

src/CloudNative.CloudEvents.Kafka/KafkaExtensions.cs
Abuntxa: Added overload for binary compatibility

jskeet:
(Just to keep communicating - this is still on my plate, but keeps being bumped by higher priority things. I promise I'll get to it eventually...)

Abuntxa:
I tried 2 different approaches for my same problem:

- Consume external dapr generated messages natively. This was considered first, as we already were consuming other non cloud event messages from kafka and didn't ourselves use dapr. It required this PR to succeed in order to be able to complete it.
- Use dapr too for consuming these external dapr generated messages.

The cloud event had extended properties with the tenant information and due to a lack of proper examples I thought it would not be possible to get a hold of that tenant information using Dapr (all examples seemed to handle the payload of the cloud event and not the whole cloud event).

I turns out that by not using the middleware added with "[.UseCloudEvents()](https://github.com/dapr/dotnet-sdk/blob/c47e65021b36c337c87d3e71c92650cb41176d82/src/Dapr.AspNetCore/DaprApplicationBuilderExtensions.cs#L54)" and parsing the cloud event directly using the [extension methods provided in this repository](https://github.com/cloudevents/sdk-csharp/blob/b4632f426db5706b85f247bf908d31be0422503e/src/CloudNative.CloudEvents.AspNetCore/HttpRequestExtensions.cs#L50) did the trick.

**Therefore, the Pull request won't be necessary** from my part, although I think that other people could find it interesting (specially for binary key support).

**Some notes regarding the dapr implementation**
During the development I missed a proper parameter binding implementation that would allow binding CloudEvents to controller method parameters. A generic CloudEvent< T > where the T would be the expected payload for deserialization could have made the trick. This would make the development easy while giving full access to the cloud envelope for accessing the custom properties.

jskeet:
Thanks so much for the update - that's really good to know. I was hoping to review this in the next couple of days, but I was getting queasy about it, knowing how little I know about Kafka.

I suggest we just leave the PR open for a little while, and see whether it attracts any attention from other users. Does that work for you?

PR #318, opened 2024-11-27
First approach of an implementation of a Pulsar protocol binding

alberto-ciklum:


jskeet:
Is the protocol documented anywhere? It would be unusual to implement a binding that isn't present in https://github.com/cloudevents/spec/tree/main/cloudevents

Directory.Packages.props
```diff
@@ -1,24 +1,24 @@
...
- 		<PackageVersion Include="Nullable" Version="1.3.1" />
- 		<PackageVersion Include="System.Memory" Version="4.5.5" />
- 		<PackageVersion Include="System.Text.Encodings.Web" Version="8.0.0" />
- 		<PackageVersion Include="System.Text.Json" Version="8.0.4" />
- 		<PackageVersion Include="xunit" Version="2.9.0" />
- 		<PackageVersion Include="xunit.runner.console" Version="2.9.0" />
- 		<PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
-	</ItemGroup>
-</Project>
+  <ItemGroup>
```
jskeet, line 2: Please revert all the whitespace changes here. By all means update the System.Text.Json dependency, but any reformatting should be done as a standalone commit (if we actually want to do it at all). That keeps the changes clear.

jskeet (commented):
Please revert the change of deleting the conformance submodule.

alberto-ciklum:
@jskeet sorry bothering you, that was al internal approach in my company. Accidentally pushed it here. I delete this PR.

Issue #321, opened 2025-04-25
Allow users to signal whether extensions are required


lailabougria:
**Use case**: From a producer-perspective, I want to avoid poison messages being produced due to missing attributes that are required in the system (either by intermediaries or consumers). This should be handled as a cross-cutting concern, to avoid gaps.

Currently, the SDK allows extensions to be created, and it allows users to provide a validator to execute.

The validator method can be used to ensure the type, structure, or even combination of extensions that should be set, providing full validation of the extension. However, it doesn't allow users to indicate that it is a required extension.

Taking the example of a `tenant` attribute that is required within the system's boundaries, I could register the attribute as follows:

```csharp
var tenantIdAttribute =
    CloudEventAttribute.CreateExtension("tenant", CloudEventAttributeType.String, tenantValue =>
    {
        if (tenantValue is not string tenant) throw new ArgumentException("tenant attribute must be of type string");

        if (string.IsNullOrWhiteSpace(tenant))
            throw new ArgumentException("tenant attribute must not be null, empty or whitespace");
    });
```

However, the validation only runs if the attribute is actually set, but from a user's perspective, the ability to indicate this as a required extension is equally important. Currently, the SDK allows me to tackle extension validation as a cross-cutting concern by providing a validation method to the SDK, however, it currently neglects what may be most essential to many users, verifying that the attribute is actually there.

jskeet:
Eek, I apologise for not responding on this - I hadn't even seen it until just now for some reason. Will respond properly tomorrow.

jskeet:
I'm torn on this. On the one hand, I can see the appeal of putting it on the extension. On the other hand, I tend to jealously guard API surfaces. It feels like you'd need to make sure you pass the attribute in wherever you're serializing/deserializing the event, so if you've centralized that logic, you could put the validation there.

As an alternative idea, how about I introduce a decorator CloudEventFormatter that could:

- Accept another CloudEventFormatter to delegate to
- Perform arbitrary validation on both serialization and deserialization (either via inheritance or passing in some sort of validator)
- Perhaps have the built-in ability to provide a set of extensions

I think that would probably still solve your concern, but also solve other validation use cases.

If that sounds like it *might* be feasible, I can whip up a prototype PR to explore API options.

lailabougria:
It sounds like that would indeed address the concern so I look forward to seeing a proposal!

jskeet:
[...]

PR #322, opened 2025-05-13
Prototype of a delegating CloudEventFormatter

jskeet:
This is the proposed (but certainly not finalized) approach for #321.

### DIFF
--- a/src/CloudNative.CloudEvents/CloudEventFormatters.cs
+++ b/src/CloudNative.CloudEvents/CloudEventFormatters.cs
@@ -1,0 +1,205 @@
+// Copyright 2025 Cloud Native Foundation.
+// Licensed under the Apache 2.0 license.
+// See LICENSE file in the project root for full license information.
+
+using CloudNative.CloudEvents.Core;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net.Mime;
+using System.Threading.Tasks;
+
+namespace CloudNative.CloudEvents;
+
+/// <summary>
+/// Extension methods for <see cref="CloudEventFormatter"/>.
+/// </summary>
+public static class CloudEventFormatters
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="formatter"></param>
+    /// <param name="extensionAttributes"></param>
+    /// <returns></returns>
+    public static CloudEventFormatter WithDefaultDecodingExtensions(this CloudEventFormatter formatter, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+        new DefaultExtensionsFormatter(formatter, extensionAttributes);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="formatter"></param>
+    /// <param name="validationAction"></param>
+    /// <returns></returns>
+    public static CloudEventFormatter WithValidation(this CloudEventFormatter formatter, Action<CloudEvent> validationAction) =>
+        WithValidation(formatter, validationAction, validationAction);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="formatter"></param>
+    /// <param name="encodingValidationAction"></param>
+    /// <param name="decodingValidationAction"></param>
+    /// <returns></returns>
+    public static CloudEventFormatter WithValidation(this CloudEventFormatter formatter, Action<CloudEvent>? encodingValidationAction, Action<CloudEvent>? decodingValidationAction) =>
+        new ValidatingFormatter(formatter, encodingValidationAction, decodingValidationAction);
+
+    /// <summary>
+    /// Delegating formatter which applies additional validation before each encoding method, and/or after each decoding method.
+    /// </summary>
+    private class ValidatingFormatter : DelegatingFormatter
+    {
+        private readonly Action<CloudEvent>? encodingValidationAction;
+        private readonly Action<CloudEvent>? decodingValidationAction;
+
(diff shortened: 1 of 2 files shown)


jskeet:
(I'd forgotten about signing-off the commit, but I haven't updated it, because we definitely don't want to merge it in the current state...)

src/CloudNative.CloudEvents/CloudEventFormatters.cs
```diff
@@ -0,0 +1,205 @@
...
+    public static CloudEventFormatter WithDefaultDecodingExtensions(this CloudEventFormatter formatter, IEnumerable<CloudEventAttribute> extensionAttributes) =>
+        new DefaultExtensionsFormatter(formatter, extensionAttributes);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <param name="formatter"></param>
+    /// <param name="validationAction"></param>
+    /// <returns></returns>
+    public static CloudEventFormatter WithValidation(this CloudEventFormatter formatter, Action<CloudEvent> validationAction) =>
```
captainsafia, line 35: I don't suppose it's strictly necessary but it looks like there's no builder pattern here. Not sure if we'd want an API that allows users to write code like:

```csharp
var formatter = new JsonEventFormatter()
  .WithRequiredTenantExtension()
  .WithEventFormatValidation()
```

src/CloudNative.CloudEvents/DelegatingFormatter.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.IO;
+using System.Net.Mime;
+using System.Threading.Tasks;
+
+namespace CloudNative.CloudEvents;
+
+// TODO:
+// - Naming:
+//   - Is the namespace appropriate? We have very few types in this root namespace
+//   - Delegating or Decorating?
```
captainsafia, line 17: I'm torn between the two but I think `Delegating` is the right way to go here. Decorating implies that data is modified/enhanced in some way whereas delegating does a better job of describing it's strictly behavioral, IMO.

src/CloudNative.CloudEvents/CloudEventFormatters.cs
jskeet: That's actually *exactly* the kind of code I'd expect them to write, to be honest. And that ends up with multi-layer delegation, but I think that's okay, given that they'll be implemented by separate classes.

Happy to understand any concerns you have though. We would want to clearly document the order in which this sort of "pipeline" is applied, of course...

captainsafia: Disregard me -- I'm blind. We're chaining on the actual `CloudEventFormatter` here. No builder required.

lailabougria:
@jskeet First of all, my apologies for taking so long. I basically dissapeared for a couple of weeks of back-to-back conferences.

I do have a couple of questions as I'm new to the code base and there aren't any tests that show how this API is to be used, so apologies if they're straightforward:
- Based on the fact that the `CloudEventFormatters` is public, my assumption is that this API is to be used by users to register validation actions?
- Does this mean the `CreateExtension` method would change, removing the validationAction from the current API?
- Currently it looks like a user would call `WithValidation` only once, and the action should contain all validations for that specific type of CloudEvent. Given that currently, validation actions are registered per extension, I was wondering what the underlying thought process was. As a user, I liked the seperation of validation action per extension as it keeps that code smaller and focussed, although the downside is that if there are cross-extension validations (validation applying to multiple extensions at a time) it becomes trickier.

jskeet:
> Based on the fact that the CloudEventFormatters is public, my assumption is that this API is to be used by users to register validation actions?

Well, to create new CloudEventFormatter instances which *do* perform validation. (It's not used to modify any existing instances.)

@captainsafia's suggestion is that the extension methods there should be the *only* public API surface, at least for the moment.

> Does this mean the CreateExtension method would change, removing the validationAction from the current API?

No, they're performing different kinds of validation. `CloudEventAttribute.CreateExtension` can specify validation that is just at the scope of "is this value valid for this extension" whereas `CloudEventFormatters.WithValidation` is intended to validate an overall CloudEvent being formatted/parsed by a CloudEventFormatter.

> Currently it looks like a user would call WithValidation only once, and the action should contain all validations for that specific type of CloudEvent.

Yes. The intention is there are two different type of validation:

- CloudEventAttribute-specific: "Is the given value valid for this CloudEvent"
- CloudEvent-wide: "Does this CloudEvent have everything I expect it to" - which would include whether "required" extensions are all present

If an extension attribute is required, that can only be validated if *some* validator knows about it - so if a CloudEventFormatter is provided with a CloudEvent which doesn't include the attribute, and asked to encode it, how could it know that the extension attribute which is missing should be present? I think that *has* to be known by the CloudEventFormatter.

For the very narrow situation of "I only care about decoding, and I only care about independently-required attributes" it would be possible to let CloudEventFormatter look at the extension attributes provided on the decode call - but that wouldn't help with cases such as "Either both X and Y must be specified, or neither" which are easliy handled by the CloudEvent-wide approach in this PR.

lailabougria:
@jskeet Thanks for clarifying, that makes sense.

I appreciate keeping these types of validations seperate, that's ideal.
I look forward to see this being included as it would help to address the use cases I had in mind.

Thank you!

jskeet:
@lailabougria Great, thanks. I'll turn this into a proper PR with documentation and tests when I have the time... but at the moment, that could be in quite a while.

Until I've got time to put this into a submittable state (unless you have the time to work on it yourself, of course) - you could potentially copy this code into your own codebase, and then just drop it when it's part of the SDK. It's only additive, and only uses public members as far as I can remember.

lailabougria:
@jskeet My summer vacation is about to start, but I will look at it when I return if you haven't had the chance to start yet.

PR #341, opened 2026-04-10
`ArrayBufferWriter` instead of `MemoryStream` for `Utf8JsonWriter`

erwinkramer:
Instead of doing `ToArray` we can do `WrittenMemory` to return the (readonly) memory, this should reduce allocation, memory and improve speed.

This won't work on netstandard2.0, the writer isn't included via `System.Buffers`... so if you like this setup then 3 methods will get different signatures for the old frameworks, which isn't elegant. This draft just displays the basic intention.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,6 +4,7 @@
     <PackageVersion Include="AMQPNetLite.Serialization" Version="2.5.1" />
     <PackageVersion Include="Apache.Avro" Version="1.12.1" />
     <PackageVersion Include="Confluent.Kafka" Version="2.14.0" />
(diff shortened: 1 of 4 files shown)


erwinkramer:
@jskeet let me know your thoughts

jskeet:
Will do when I can, but I don't know when I'll get to it. Maybe over the weekend, but I'm not promising.

jskeet:
Given that this one looks pretty fiddly, I think it would be good to compare performance. If it's only a small part of the performance penalty, it may not be worth it... if it makes a *huge* difference, it's worth looking at. Having some benchmarks for the project would be a good thing in general...

PR #349, opened 2026-05-05
Cover the non transcoding stream option in the SystemTextJson JsonEventFormatter unit test

erwinkramer:
This PR makes sure to cover the non-transcoding path.

This is a follow up of https://github.com/cloudevents/sdk-csharp/pull/342

I considered targeting `netstandard2.0` for the unit tests but that seemed to be much more work and out of proportions for this scenario.

### DIFF
--- a/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
+++ b/src/CloudNative.CloudEvents.SystemTextJson/JsonEventFormatter.cs
@@ -104,6 +104,12 @@ public class JsonEventFormatter : CloudEventFormatter
     protected JsonDocumentOptions DocumentOptions { get; }
 
     /// <summary>
+    /// Whether non-UTF8 JSON should be transcoded to UTF-8 before parsing when the target framework supports it.
+    /// Derived classes may override this to exercise the fallback path.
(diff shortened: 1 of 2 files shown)


jskeet:
I'm going to sit on this for a while - because I think there's value in targeting .NET 4.8 (using a Windows runner) in general. That would give me more confidence that the whole set of libraries works in .NET 4.8 than just relying on this "make sure we test this code path" approach. But if I can't get that to work, we can fall back on this.

jskeet:
#353 is my alternative to this, which covers the non-transcoding path more naturally. We'd then want to make the .NET Framework tests run in GitHub, but that can be done separately.

erwinkramer:
Yeah, that's better for the use-case. I guess this current PR now only makes sense if we want to publicly expose the option of `UseTranscodingStreamForNonUtf8` for .net8 (and up) users (but i doubt that makes sense). 

I'm happy to close this PR unless you see otherwise.

jskeet:
Closing this PR works for me. If you could give an LGTM to #353 (I'm not sure whether the thumbs-up is "LGTM" or "yes I'll review" at the moment) I'll merge that.

PR #351, opened 2026-05-05
AMQPNetLite.Core instead of AMQPNetLite

erwinkramer:
This PR reduces the dependency of AMQPNetLite to just AMQPNetLite.Core

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,7 +1,6 @@
 <Project>
   <ItemGroup>
-    <PackageVersion Include="AMQPNetLite" Version="2.5.2" />
-    <PackageVersion Include="AMQPNetLite.Serialization" Version="2.5.2" />
+    <PackageVersion Include="AMQPNetLite.Core" Version="2.5.2" />
     <PackageVersion Include="Apache.Avro" Version="1.12.1" />
     <PackageVersion Include="Confluent.Kafka" Version="2.14.0" />
     <PackageVersion Include="Google.Protobuf" Version="3.34.1" />
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -13,7 +13,7 @@ unnecessary dependencies. These packages are:
 |NuGet package|Description|
 |-|-|
 |[CloudNative.CloudEvents](https://www.nuget.org/packages/CloudNative.CloudEvents)|Core SDK|
-|[CloudNative.CloudEvents.Amqp](https://www.nuget.org/packages/CloudNative.CloudEvents.Amqp)|AMQP protocol binding using [AMQPNetLite](https://www.nuget.org/packages/AMQPNetLite)|
+|[CloudNative.CloudEvents.Amqp](https://www.nuget.org/packages/CloudNative.CloudEvents.Amqp)|AMQP protocol binding using [AMQPNetLite.Core](https://www.nuget.org/packages/AMQPNetLite.Core)|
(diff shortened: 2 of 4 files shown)


erwinkramer:
The build error seems unrelated to the PR and improved via https://github.com/cloudevents/sdk-csharp/pull/352

jskeet:
It's unclear to me how the two packages relate to each other - in particular, which assembly `Message` belongs to when you depend on `AMQPNetLite`. This is either a "just about breaking change" for customers who are relying on the dependency on AMQPNetLite for types that aren't in AMQPNetLite.Core, or "definitely a binary breaking change" if we're changing which assembly things come from.

I'll need to take a bit of time to look into this more carefully - not sure when that will be, I'm afraid. (I'm going to prioritize the "get tests running in .NET 4.8".)

erwinkramer:
I checked the assemblies. The signature is the same: `Assembly Amqp.Net, Version=2.5.0.0, Culture=neutral, PublicKeyToken=905a7b1e6458e0c3`, so I don't see how it could cause an issue:

Original:

```csharp
#region Assembly Amqp.Net, Version=2.5.0.0, Culture=neutral, PublicKeyToken=905a7b1e6458e0c3
// C:\Users\x\.nuget\packages\amqpnetlite\2.5.2\lib\net5.0\Amqp.Net.dll
// Decompiled with ICSharpCode.Decompiler 9.1.0.7988
#endregion

using System;
using Amqp.Framing;
using Amqp.Types;

namespace Amqp;

//
// Summary:
//     The Message class represents an AMQP message.
public class Message : IDisposable
{

...
```

Core:

```csharp
#region Assembly Amqp.Net, Version=2.5.0.0, Culture=neutral, PublicKeyToken=905a7b1e6458e0c3
// C:\Users\x\.nuget\packages\amqpnetlite.core\2.5.2\lib\netstandard2.0\Amqp.Net.dll
// Decompiled with ICSharpCode.Decompiler 9.1.0.7988
#endregion

using System;
using Amqp.Framing;
using Amqp.Types;

namespace Amqp;

//
// Summary:
//     The Message class represents an AMQP message.
public class Message : IDisposable
{

...
```

jskeet:
I definitely want to check. Example of how this could get weird - suppose someone already has a dependency on `AMQPNetLite`. Presumably that package will be installed *as well as* `AMQPNetLite.Core`. Does that cause problems? Hopefully not, but I'd want to check to make sure. What happens if those two packages are installed at different versions? Which assembly will be used? (Overall, it feels like an antipattern for the same type to be in two different packages - making this sort of change risky.)

erwinkramer:
Yeah, let me know your findings. They have some docs here explaining a little bit about the positioning of the libraries: https://azure.github.io/amqpnetlite/articles/installation.html#nuget-packages

jskeet:
I won't have time to do as thorough an investigation into this as I'd want, but I *can* confirm that this works:

- TestAmqpLib depends on AMQPNetLite and AMQPNetLite.Serialization
- TestAmqpApp depends on TestAmqpLib and CloudNative.CloudEvents.Amqp, and uses Message

So before the change, there's no discrepancy at all, but after the change the two libraries are using different NuGet packages (but the same assembly name). Fortunately, it still builds with no issues at all.

My temptation is to just merge this without bumping the major version, but be ready to be versatile after the next release if users report issues. Does that sound reasonable to you?

erwinkramer:
> Does that sound reasonable to you?

Sounds good to me.

PR #352, opened 2026-05-05
Race condition handling on httplistener in unit tests

erwinkramer:
As we see here https://github.com/cloudevents/sdk-csharp/actions/runs/25399467082/job/74494893548?pr=351#step:4:102, this looks like a pretty rare race condition. 

This PR catches the `HttpListenerException` and tries a few more times with a new random port. That hopefully should solve it.


jskeet (approved):
Great, thanks :)

PR #353, opened 2026-05-12
Make unit tests run on .NET 4.8 on Windows.

jskeet:
This doesn't run the tests in CI yet, but it's a start.

@erwinkramer are you happy to review this? The changes are basically the least I could do to make it all work in .NET 4.8 :)

PR #355, opened 2026-05-18
Run workflow on Windows too; and when on Windows, run the net48 tests explicitly

erwinkramer:
Succeeding https://github.com/cloudevents/sdk-csharp/pull/353.

I made the test runners explicitly target net8, net10 and net48, because https://github.com/actions/setup-dotnet doesn't support net48. This causes them to not run parallel anymore, but I couldn't see an elegant solution to keep them parallel. However, both Linux and windows run in parallel now, and Windows is slower than Linux now anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,16 @@ on:
   pull_request:
 
 jobs:
-
   build:
-    runs-on: ubuntu-24.04
+    strategy:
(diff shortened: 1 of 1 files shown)


jskeet (approved):
Looks good to me. I'm not too worried about not running things in parallel - it's not a hugely slow test set anyway.

PR #357, opened 2026-05-18
`net10` target and drop redundant `netstandard2.1` target

erwinkramer:
Added net10 as target in the core projects. I removed netstandard2.1 as I don't see how that adds any value, it's overlapping with netstandard2.0, net8 and net10.

### DIFF
--- a/src/CloudNative.CloudEvents.Amqp/CloudNative.CloudEvents.Amqp.csproj
+++ b/src/CloudNative.CloudEvents.Amqp/CloudNative.CloudEvents.Amqp.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netstandard2.1;net8.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net8.0;net10.0</TargetFrameworks>
     <Description>AMQP extensions for CloudNative.CloudEvents</Description>
     <Nullable>enable</Nullable>
     <PackageTags>cncf;cloudnative;cloudevents;events;amqp</PackageTags>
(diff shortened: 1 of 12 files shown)


jskeet:
I think the value of netstandard2.1 is if the project is consumed by any *other* netstandard2.1 libraries - otherwise they'll get the netstandard2.0 build, I believe, and that might incur more dependencies. I agree it's all a bit unfortunate though.

Is there a particular benefit in adding a net10.0 target, rather than just letting net10.0 projects consume net8.0?

erwinkramer:
For .NET 10, I don't think we benefit from that right now in a technical sense, but when we do want to add specific features for .net 10, on a project level or in the code itself, then we're ready for it. 

In general, it might clear up the library intention: 
1. support net48 with netstandard2.0, and 
1. directly support net8 and net10.  

In theory that's all it has to support, right? Then, when net8 reaches End-Of-Support later this year, we can drop that one to make it clear we're just on the Microsoft-supported targets. I know there are edge cases, but they should be formally out of support by Microsoft, and thus by this library as well, at least that's my reasoning.

jskeet:
Sorry for not replying before. I'm finally getting round to prepping for a release, and figure it would be good to decide on this PR beforehand. (Having missed this, and as it needs consideration, that may well delay the release a bit while we discuss it, but I think that's okay.)

I do think we still need to *consider* other libraries targeting netstandard2.1. If we remove the target for that, any consumers targeting netstandard2.1 will end up with the System.Memory dependency (because they'll end up using the netstandard2.0 target). I can imagine that *potentially* causing issues... but equally maybe we can say it's okay. I think it's probably worth deciding once and for all though, so we don't need to revisit this every year or so.

We could reasonably assume that a) not many libraries target netstandard2.1; b) those that do will just have to cope; c) we could always reintroduce a netstandard2.1 target if it turns out it creates really significant issues. I think I could live with that.

If we go with that, I think this PR is ready to go. How do you feel about that reasoning?

erwinkramer:
Sounds good to me.

jskeet (approved):

PR #366, opened 2026-07-04
Update validation target versions.

jskeet:
(This is always the first task after releasing a new major or minor version.)


jskeet:
(Again, will self-merge this PR on green.)

PR #374, opened 2026-08-19
Include specversion in GetPopulatedAttributes

### LINKED ISSUE #373: Include specversion in CloudEvent.GetPopulatedAttributes
## Summary

`CloudEvent.GetPopulatedAttributes()` currently excludes `specversion`, even though the spec version is always populated on a `CloudEvent`. Callers that enumerate context attributes therefore omit a required CloudEvents attribute unless they special-case `CloudEvent.SpecVersion.VersionId`.

## Cross-SDK behavior

The Java SDK includes `specversion` in its equivalent enumeration:

- `SpecVersion.V1.getAllAttributes()` includes `specversion`.
- `CloudEventAttributes.getAttributeNames()` filters that collection to non-null values.
- `CloudEventImplTest.testGetAttributeNames()` explicitly expects `specversion`.

This difference surfaced while mapping a C# `CloudEvent` to protocol properties by iterating `GetPopulatedAttributes()`; the resulting event had the other populated attributes but no `specversion`.

## Proposed behavior

Return `CloudEventsSpecVersion.SpecVersionAttribute` with `SpecVersion.VersionId` from `GetPopulatedAttributes()`, alongside other populated context and extension attributes.

The built-in formatters and bindings already serialize spec version separately, so they should skip this pair during their existing enumeration to keep wire output unchanged.

I intend to submit a PR with the implementation, regression test, and updates to internal enumeration call sites.

authuir:
## Summary

- include `specversion` and its canonical value in `CloudEvent.GetPopulatedAttributes()`
- keep existing AMQP, HTTP, Kafka, Avro, JSON, and Protobuf wire output unchanged by skipping the version where it is already serialized separately
- add a regression test and update the test event clone helper for the read-only version attribute

This aligns attribute enumeration with the Java SDK, where `getAttributeNames()` includes the populated `specversion` context attribute.

Fixes #373

## Testing

- `git diff --check`
- VS Code C# diagnostics for all changed files

The focused `CloudEventTest` run could not start locally because NuGet restore repeatedly timed out connecting to `api.nuget.org` (SSL transport error: socket not connected).


jskeet:
As it stands, this is a breaking change - the method is explicitly documented not to include the version attribute. The version attribute is "special" in various other ways - for example, it's read-only, and you can't access it via the indexer.

If you really, really think there ought to be a way of obtaining all the attributes including the version (which is frankly easy enough to do at the caller side where necessary) we could introduce a new overload:

```csharp
public IEnumerable<KeyValuePair<CloudEventAttribute, object>> GetPopulatedAttributes(bool includeSpecVersionAttribute)
```

The existing method could just return `GetPopulatedAttributes(false)`. That would also mean that none of the other code needs to change.

Personally I'm not really convinced it's worth doing this, but that would at least be *much* better than a breaking change.

jskeet:
Closing this PR due to lack of engagement.

